PR #9, opened 2017-07-21
Centralize GraphQL queries

rschlaefli:
GENERAL
- Move all queries from `TagList`, `QuestionList` and `SessionList` to a centralized file `queries.js`
- Prepare a file for mutations `mutations.js`
- Prepare queries (drafts) for several components that are not yet implemented


rschlaefli:
[GraphQL Queries erarbeiten](https://trello.com/c/uOYYabMo/270-graphql-queries-erarbeiten)

rschlaefli:
Will be merged after pending review.

src/queries/queries.js
```diff
@@ -0,0 +1,92 @@
+import { gql } from 'react-apollo'
```
pascalzehnder: How shall we sort these queries? Did you sort them according to categories? As I see you didn't sort them in alphabetical order. Maybe we should name them as their name express the component in which it is used.

src/queries/queries.js
```diff
@@ -0,0 +1,92 @@
...
+        versions {
+          id
+          createdAt
+        }
+        createdAt
+        updatedAt
+      }
+  }
+`
+const QrGeneratorQuery = gql`
```
pascalzehnder: According to my first comment: To which component does this query belong?

pascalzehnder (approved):
I like having the GraphQL queries in a centralised file. As the amount of the queries is increasing, we need to pay attention having a predefined order of the queries.

PR #15, opened 2017-08-11
Upgrade node image version to 8.3.0

rschlaefli:


rschlaefli:
The image will need to be checked. Is it still runnable?

PR #22, opened 2017-08-14
Integrate react-helmet

rschlaefli:
**GENERAL**
- Integrate react-helmet in base next `_document` (server-side rendering)
- Update `withCSS` to make use of react-helmet instead of next/head
  - This allows for deduplication of links in head
- Integrate a `pageTitle` prop in layouts and pass it from all pages
- Update flow types for head prop (next/head to any)

**DEPENDENCIES**
- Install react-helmet

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "react-addons-css-transition-group": "^15.6.0",
     "react-apollo": "^1.4.3",
     "react-dom": "^15.6.1",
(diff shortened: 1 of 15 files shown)


rschlaefli:
@pascalzehnder This would allow to use `withCSS` on a component level and still not have any duplicate link tags in the document head.

PR #24, opened 2017-08-14
Rework inclusion of <link> components in <head>

rschlaefli:
**GENERAL**
- Remove overly complicated `WithCSS` HOC and all usages in components and pages
  - No longer need to include a head prop and insert {head} into the component tree
  - Instead use `<Helmet>{createLinks([...full urls], [...semantic components])}</Helmet>` inside the component
- Add utility function for expanding list of strings to <link href> elements
- Make use of helmet everywhere `WithCSS` was used before

**MISC**
- (fix) Many type errors that started occurring "randomly" (but not in respective PR #23)


rschlaefli:
@pascalzehnder Bitte review, aber ohne die ganzen Typing-Fehler Commits.

src/pages/questions/index.js
```diff
@@ -79,7 +79,7 @@ class Index extends Component {
       }),
     }
 
-    const actionButton = (
+    const actionButton: any = (
```
rschlaefli: This was necessary to make the flow typechecks run. Why??

PR #41, opened 2017-08-18
Simplify renovate.json

rarkins:


renovate.json
```diff
@@ -1,10 +1,6 @@
 {
-  "extends": [":library"],
+  "extends": [":app"],
```
rarkins: Renovate deduced that you were using a non-public package.json (i.e. you haven't set `private=true`) so therefore defaults to the `:library` preset. `:library` is essentially the same as `:app` except that it retains semver ranges for `dependencies`, because it's considered good form for libraries to not pin dependencies and risk causing duplicates in downstream packages.

Anyway I have changed this to `:app` which pins everything, so there's then no need for `depTypes` version pinning added manually now.

rarkins:
By the way, your choice of groupings is interesting. Let me know if your opinions on how to group/merge them evolve over time.

My own experience and it seems to be everybody's is initial happiness at the automation ("yes! another PR.. how efficient") but after a while you wish for less. Grouping and scheduling and two ways to achieve this, but automerging is another one for you to consider. For example, if any test dependencies are upgraded and your CI passes al tests.. are you really going to do any more investigation before clicking merge? Hence why I'm considering classifying all test-related packages into a preset-group that has automerge set to `branch-push`, i.e. skip PRs unless tests fail.

rschlaefli:
Hi @rarkins, many thanks for your suggestions and improvements! 

I just grouped the dependencies as I thought might fit well, as I don't really want to be overloaded with pull requests. Of course I will have to see how it all plays out. Automerging for testing and/or code style dependencies sounds like a very nice feature. I will have a look into it, thanks!

PR #81, opened 2017-09-13
Creating Free-Text Form

pascalzehnder:
- Add specific .DS_Store file to gitignore
- Add 'Free Text' as a choice to question creation form
- Add specific free text preview including displaying slider which simulates the real slider the students have in their screen
- Add Possibility setting a minimum and/or a maximum to a number range
- Create Stories


src/components/forms/QuestionCreationForm.js
```diff
@@ -84,15 +88,24 @@ const QuestionCreationForm = ({
...
+      <Field
+        name="content"
+        component={type === 'SC' || type === 'MC' ? SCCreationContent : SCCreationContent}
+      />
     </div>
 
-    <div className="questionInput questionOptions">
-      <Field name="options" component={SCCreationOptions} />
-    </div>
+    {type !== 'FREE' && (
```
rschlaefli: `<div className="questionInput....` should wrap the different options (don't contain inside {type !== ...}

src/components/forms/QuestionCreationForm.js
```diff
@@ -84,15 +88,24 @@ const QuestionCreationForm = ({
...
+        <Field name="options" component={SCCreationOptions} />
+      </div>
+    )}
 
     <div className="questionPreview">
-      <SCCreationPreview title={title} description={content} options={options} />
+      {type === 'SC' || type === 'MC' ? (
+        <SCCreationPreview title={title} description={content} options={options} />
+      ) : (
+        <FREECreationPreview title={title} description={content} />
```
rschlaefli: There would be two previes components depending on the chosen restrictions.

src/components/questionTypes/Free/creation/Content.js
```diff
@@ -0,0 +1,29 @@
+// @flow
```
rschlaefli: This component will be the same as for the other question types, we can put this in commons.

src/components/questionTypes/Free/creation/Preview.js
```diff
@@ -0,0 +1,98 @@
...
+  },
+  title: 'TITLE',
+}
+
+const Preview = ({ title, options, description }: Props) => (
+  <div className="preview">
+    <div className="title">{title || 'TITLE'}</div>
+    <div className="description">{description || 'DESCRIPTION'}</div>
+    {
+      options.restrictions.type === 'NUMBERS' && options.restrictions.min && options.restrictions.max &&
```
rschlaefli: @pascalzehnder If type is NUMBERS and min and max are not set, display a simple input field for an arbitrary number (basically another type of restriction).

src/components/questionTypes/FREE/creation/options/Options.js
```diff
@@ -0,0 +1,132 @@
...
+        type,
+      },
+    })
+  }
+
+  render() {
+    const { intl, input: { value } } = this.props
+
+    // TODO: extract to FormattedMessage
+    const optionsData = [
```
rschlaefli: Extract to FormattedMessages

src/components/questionTypes/FREE/creation/options/Options.js
```diff
@@ -0,0 +1,132 @@
...
+
+import React from 'react'
+import classNames from 'classnames'
+import { FormattedMessage } from 'react-intl'
+
+import type { TextInputType } from '../../../../../types'
+
+type Props = {
+  intl: $IntlShape,
+  input: TextInputType,
```
rschlaefli: TextInputType is wrong for this kind of component (will result in flow errors). It expects value to be a string and onChange to update this string. You will need to write something like:

`input: {
  value: {
    restrictions: {
      type...
      min....
    }
  },
  onChange: (any) => void,
}`

PR #146, opened 2017-10-16
Student View Optimisations

pascalzehnder:
- Implementation of adding a new feedback
- Visualisation of different answer options depending on question type
- console.log when sending an answer
- adding a new feedback then append value to state with recompose
- up-voting of feedbacks in state
- if user already up-voted an item then the user can't update it anymore


rschlaefli:
**DEPENDENCIES**
- Install `js-cookie` for clientside cookie management

**GENERAL**
- Add dynamic route with new page at `/join/:shortname` for joining a running session
- Implement responding to all different question types
- Improve layout and page structure for student views

**MISC**
- Small improvements on evaluation page
- (fix) Session creation
- (fix) Creation of MC questions
- (fix) Validation on registration form
- Slight layout adjustments for question blocks
- Add *.graphql files for all queries and mutations as a preparatory step

PR #203, opened 2017-11-14
Histogram

pascalzehnder:
Create a histogram for evaluation of FREE question within a number range

**GENERAL**
- Create New Histogram returning a recharts visualisation.
- mapping data to correct format including the value and the count of the value.


pascalzehnder:
[Komponente: HistChart](https://trello.com/c/LhGzYcif/430-komponente-histchart)

rschlaefli:
**DEPENDENCIES**
- Update recharts from `1.0.0-beta.1` to `1.0.0-beta.2`

**GENERAL**
- Improve structure of mapping logic for active instance data
- Extract statistics into separate component and relayout using `<List>`
- Rework statistics calculations using recompose and map/reduce and move to separate tested library file ``math.js`
- Improve histogram component
  - Include mean and median
  - Use "quasi-binning" with a bin-size of 1 to get to integer indices (make it work with float responses)
  - Optionally include the solution (not yet in the API)
- Optimizations on the evaluation page layout
  - Always use the entire screen height
  - Improve chart margins

**MISC**
- Add story for statistics component
- Don't show the `<Brush>` if the session is running, as refetches will reset the slider state

@pascalzehnder LGTM :)

PR #407, opened 2018-03-05
Pie and Bar Chart Fix

AlexScheitlin:
Fixed pie and bar chart labels for too small pies and bars.

Depending on a threshold (manually set) the charts decide whether to display the label within or outside the pie/bar.


src/components/evaluation/charts/BarChart.js
```diff
@@ -49,6 +49,25 @@ const calculatePercentage = (questionType, count, totalResponses) => {
   return ''
 }
 
+// break point for too small bars
+// if the percentual responses of a bar are smaller than the given
+// value, the label (A, B, ...)  is not displayed on top of the bar
+const smallBarBreak = 0.05
```
rschlaefli: Should be defined as a constant in the constants file

src/components/evaluation/charts/BarChart.js
```diff
@@ -49,6 +49,25 @@ const calculatePercentage = (questionType, count, totalResponses) => {
...
+
+const getPercentageIn = (questionType, count, totalResponses) => {
```
rschlaefli: needs some comments

src/components/evaluation/charts/BarChart.js
```diff
@@ -49,6 +49,25 @@ const calculatePercentage = (questionType, count, totalResponses) => {
...
+  if (count / totalResponses > smallBarBreak) {
+    return calculatePercentage(questionType, count, totalResponses)
+  }
+  return ''
+}
+
+const getPercentageOut = (questionType, count, totalResponses) => {
```
rschlaefli: needs some comments

src/components/evaluation/charts/PieChart.js
```diff
@@ -72,14 +72,42 @@ const PieChart = ({ isSolutionShown, data }) => (
 PieChart.propTypes = propTypes
 PieChart.defaultProps = defaultProps
 
+const indexToLetter = index =>
```
rschlaefli: please extract these functions into a separate file (lib/charts.js). also the functions in the bar chart.. they could be reusable for something else, so we should extract them.

src/components/evaluation/charts/BarChart.js
AlexScheitlin: Done. I also extracted the variable for too small pies to the constant file. The new names are SMALL_BAR_THRESHOLD and SMALL_PIE_THRESHOLD.

src/components/evaluation/charts/BarChart.js
AlexScheitlin: Done

src/components/evaluation/charts/BarChart.js
AlexScheitlin: Done

src/components/evaluation/charts/PieChart.js
AlexScheitlin: Done :)

PR #915, opened 2019-03-28
Added "Cancel Session" functionality (frontend)

wuergrob:
Added functionality in accordance to: https://github.com/uzh-bf/klicker-uzh/issues/12

- Cancel button is located on the running session page
- Pressing it will open a modal window, asking the user to confirm the cancellation
- Cancelling will revert the session back to its "Created" state
- Cancelling will also revert any changes made while it was running

**MISC**
- Please not that i ran into problems some of the ESlint rules. While i fixed most of them, a few remained 
for which i've added exceptions instead of properly dealing with them.

- This added funcionality will not work without the corresponding changes made to the Backend of the klicker software.


src/components/sessions/CancelModal.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable react/jsx-sort-props */
+/* eslint-disable react/destructuring-assignment */
+import React, { Component } from 'react'
+import { Button, Header, Icon, Modal } from 'semantic-ui-react'
+import { FormattedMessage } from 'react-intl'
+
+export default class CancelModal extends Component {
```
rschlaefli: TODO @rschlaefli: Replace with function and useState hook

src/components/sessions/CancelModal.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable react/jsx-sort-props */
```
rschlaefli: TODO @rschlaefli: I need to look at the lint rules in general, so we don't have to work around them like this.

src/graphql/mutations/CancelSessionMutation.graphql
```diff
@@ -0,0 +1,6 @@
+mutation CancelSession($id: ID!) {
+  cancelSession(id: $id) {
```
rschlaefli: TODO @rschlaefli: This should probably return more than just the status (returns a Session)

PR #919, opened 2019-04-01
Export evaluation in PDF format #89

mnmldev:
User should be able to export a PDF file via a button on the evaluation page, which contains the question, its answer possibilities and the default graph for all questions.

**DEPENDENCIES**
- react-pdf [4.0.5] (https://www.npmjs.com/package/react-pdf) (https://react-pdf.org/)

**GENERAL**
- uzh-bf/klicker-uzh#89


rschlaefli:
@mnmldev I found that to create PDFs, you would actually need to use https://github.com/diegomura/react-pdf. The `react-pdf` you installed is actually a display layer for PDFs.

rschlaefli:
Closing this PR as the work has moved to #1018 . Thanks for the contribution :)

PR #920, opened 2019-04-12
Reset question block cache delete

Jokkis80:
Deletes the Cache to reset a Question Block

**GENERAL**
- Deletes Redis Cache for Question Blocks


src/pages/sessions/running.js
```diff
@@ -80,60 +93,72 @@ const Running = ({ intl, shortname }) => (
...
-                                  })
-                                }}
-                                intl={intl}
-                                isEvaluationPublic={settings.isEvaluationPublic}
-                                runtime={runtime}
-                                sessionId={id}
-                                shortname={shortname}
-                                startedAt={dayjs(startedAt).format('HH:mm:ss')}
-                              />
+                                      Router.push('/sessions/running')
```
rschlaefli: There should be no redirect after a session block is reset. The necessary state should always be updated by Apollo automatically. If this is here for a state refresh, we will need to look at it in more detail.

Jokkis80: the redirect has been removed

PR #948, opened 2019-05-10
Automatic Session Update

Jokkis80:
Updates the Student View automatically

### General
session subscription implemented


src/pages/join.js
```diff
@@ -70,98 +76,120 @@ const Join = ({
...
+          }}
+          subscribeToMore={() =>
+            subscribeToMore({
+              document: UpdateSessionSubscription,
+              updateQuery: (prev, { subscriptionData }) => {
+                if (!subscriptionData.data) return prev
+                const newFeedItem = subscriptionData.data.sessionUpdateAdded
+                return Object.assign({}, prev, {
+                  joinSession: {
+                    activeInstances: [newFeedItem.activeInstances, ...prev.joinSession.activeInstances],
```
rschlaefli: I am not sure why you concatenate with the previously active instances here?

src/pages/join.js
```diff
@@ -70,98 +76,120 @@ const Join = ({
...
+                    feedbacks: [newFeedItem.feedbacks, ...prev.joinSession.feedbacks],
+                    id: [newFeedItem.id, ...prev.joinSession.id],
```
rschlaefli: What is this array of ids for?

src/pages/join.js
```diff
@@ -70,98 +76,120 @@ const Join = ({
...
+          subscribeToMore={() =>
+            subscribeToMore({
+              document: UpdateSessionSubscription,
+              updateQuery: (prev, { subscriptionData }) => {
+                if (!subscriptionData.data) return prev
+                const newFeedItem = subscriptionData.data.sessionUpdateAdded
+                return Object.assign({}, prev, {
+                  joinSession: {
+                    activeInstances: [newFeedItem.activeInstances, ...prev.joinSession.activeInstances],
+                    feedbacks: [newFeedItem.feedbacks, ...prev.joinSession.feedbacks],
```
rschlaefli: The feedbacks should never be changed by the updated session state, right? I think this could be removed.

src/pages/join.js
```diff
@@ -70,98 +76,120 @@ const Join = ({
...
+                    id: [newFeedItem.id, ...prev.joinSession.id],
+                    settings: [newFeedItem.settings, ...prev.joinSession.settings],
```
rschlaefli: The settings should also not change in a session update, as I do not think you have added a publish to the session settings?

src/pages/join.js
rschlaefli: I think if we update the session, we want to remove the previously active ones and add the new ones. But as it worked in your demo, I am not sure about the behavior of this logic here.

PR #949, opened 2019-05-10
#106 Mark Feedback as answered

schuepbach:
Mark feedback as answered to put them on the bottom of the feedback channel. (frontend)

### General

https://github.com/uzh-bf/klicker-uzh/issues/106

### DIFF
--- a/src/pages/sessions/running.js
+++ b/src/pages/sessions/running.js
@@ -18,6 +18,7 @@ import {
   EndSessionMutation,
   PauseSessionMutation,
   CancelSessionMutation,
+  UpdateFeedbackMutation,
   UpdateSessionSettingsMutation,
   ActivateNextBlockMutation,
   DeleteFeedbackMutation,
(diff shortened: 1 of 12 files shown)


src/pages/sessions/running.js
```diff
@@ -229,55 +230,85 @@ const Running = ({ intl, shortname }) => (
...
+                              })
+                            }}
+                            handleUpdateFeedback={(newTags, feedbackId) => {
+                              updateFeedback({
+                                refetchQueries: [{ query: RunningSessionQuery }],
+                                variables: {
+                                  feedbackId,
+                                  sessionId: id,
+                                  tags: newTags,
+                                  settings: {
```
rschlaefli: Why do we need to pass these settings? I think these should not change if we only update a single feedback?

src/pages/sessions/running.js
```diff
@@ -229,55 +230,85 @@ const Running = ({ intl, shortname }) => (
...
+                            subscribeToMore={() => {
+                              subscribeToMore({
+                                document: FeedbackAddedSubscription,
+                                updateQuery: (prev, { subscriptionData }) => {
+                                  if (!subscriptionData.data) return prev
+                                  const newFeedbacks = [
+                                    ...prev.runningSession.feedbacks,
+                                    subscriptionData.data.feedbackAdded,
+                                  ]
+                                  const sortedFeedbacks = newFeedbacks.sort((f1, f2) => {
```
rschlaefli: Just use this sorting logic for all the feedbacks instead of also sorting in the API, as I mentioned in the PR there :)

PR #1549, opened 2020-08-05
Finished including <FormattedMessage> in Toast in order to enable translation

lulocher:


rschlaefli (commented):
Thanks a lot! I just updated the en.json language definitions. Remaining things we should do before merge: Add short description text to the messages that only contain `{errorMessage}` and translate the strings and add them to `de.json`.

```
  "components.questions.actionBar.data.error": "{errorMessage}",
  "components.questions.actionBar.export.error": "{errorMessage}",
  "components.questions.uploadModal.load.error": "{errorMessage}",
  "components.questions.uploadModal.sequence.error": "{errorMessage}",
```

PR #1550, opened 2020-08-05
Handled overflow in in evaluation window

lulocher:


rschlaefli (changes requested):
Thanks!

Small comments:

1. Would prefer to use `overflow-y: auto` over `scroll`, as `auto` will only show a scroll bar on the right if the content needs to be scrolled (not if it is small enough to fit).
2. I would make it so that only the rows of the `TableChart` scroll, not the entire table or even any of the other charts. So, the rows would show a scrollbar, but the header of the table would always remain visible.

PR #1584, opened 2020-08-21
WIP: Migrate to formik2

lulocher:
Migrated forms to formik2 (useFormik hook)


rschlaefli (commented):
Need to think about this again :)

PR #1604, opened 2020-08-26
Disabled the complete storage mode

lulocher:
Storage option can not be chosen anymore, the corresponding modal is hidden now . 'SECRET' mode is always active until 'COMPLETE' mode is fully implemented.


rschlaefli (approved):
Great, thanks a lot!

PR #1605, opened 2020-08-27
Runtime calculator frontend

lulocher:
Removed runtime field from all requests which formerly contained it and introduced the runtime calculation in the SessionTimeline component.  Additionally added "moment" module to dependencies.


src/components/sessions/SessionTimeline.tsx
```diff
@@ -132,12 +143,24 @@ function SessionTimeline({
 
   const isFeedbackSession = blocks.length === 0
 
+  const [runtime, setRuntime] = useState(calculateRuntime({ startedAt }))
+
+  const startingTime = runtime.includes('d')
+    ? dayjs(startedAt).format('DD.MM. HH:mm:ss')
+    : dayjs(startedAt).format('HH:mm:ss')
+
+  useEffect(() => {
```
rschlaefli, line 152: Should add a cleanup handler as well (i.e., return a function that will be called when the component is unmounted).

E.g., ```
var handle = setInterval(...)
return () => clearInterval(handle)
```

File: src/components/sessions/SessionTimeline.tsx [hunk 2]
```diff
@@ -67,6 +69,19 @@ function getMessage(intl, num: number, max: number): any {
   }
 }
 
+const calculateRuntime = ({ startedAt }): string => {
+  const duration = moment.duration(moment().diff(startedAt))
```
rschlaefli: Use dayjs here instead of installing the "big" moment library for this single line of code. Dayjs can do exactly the same and is already installed as a dep (and you have used it below as well) ;)

src/components/sessions/SessionTimeline.tsx
```diff
@@ -132,12 +143,24 @@ function SessionTimeline({
 
   const isFeedbackSession = blocks.length === 0
 
+  const [runtime, setRuntime] = useState(calculateRuntime({ startedAt }))
+
+  const startingTime = runtime.includes('d')
```
rschlaefli, line 148: Great, I like having different formats when started on another day!

rschlaefli (changes requested):

src/components/sessions/SessionTimeline.tsx
lulocher: Thanks for reviewing, implemented it as suggested.

src/components/sessions/SessionTimeline.tsx
lulocher: Makes sense, changed to dayjs library now and removed the moment library from the dependencies.

src/components/sessions/SessionTimeline.tsx
lulocher: Thanks!

rschlaefli (approved):
Great, thanks!

PR #1606, opened 2020-08-27
Live Answer-Counter on Running Session Page

lulocher:
The number of answers of a running session are now displayed and automatically updated on the running session page.


src/components/sessions/SessionTimeline.tsx
```diff
@@ -124,12 +127,24 @@ function SessionTimeline({
...
   handleActivateBlockById,
+  handleActiveBlock,
+  handleNoActiveBlock,
   subscribeToMore,
 }: Props): React.ReactElement {
   useEffect((): void => {
     subscribeToMore()
   }, [])
 
+  const [isBlockActive, setIsBlockActive] = useState(activeStep % 2 === 1)
```
rschlaefli: I think you can just use a normal const for such "derived" properties that are never updated from within the component (i.e., just in your useEffect). For example

```suggestion
  const isBlockActive = activeStep % 2 === 1
```

rschlaefli: It will be recomputed on every render but that can be neglected for such trivial calculations.

src/components/sessions/SessionTimeline.tsx
```diff
@@ -124,12 +127,24 @@ function SessionTimeline({
...
+
+  useEffect(() => {
+    setIsBlockActive(activeStep % 2 === 1)
+  }, [activeStep])
+
+  useEffect(() => {
```
rschlaefli: I think you could combine everything and rewrite it like:

```javascript
useEffect(() => {
	const isBlockActive = activeStep % 2 === 1
	if (isBlockActive) {
		handleActiveBlock()
	} else {
		handleNoActiveBlock()
	}
}, [activeStep])
```

rschlaefli (changes requested):
Just some minor changes, otherwise LGTM, thanks a lot!

src/components/sessions/SessionTimeline.tsx
rschlaefli: See comment below.

src/components/sessions/SessionTimeline.tsx
lulocher: Thanks for reviewing, totally makes sense! Implemented it as suggested...

src/components/sessions/SessionTimeline.tsx
```diff
@@ -124,12 +127,25 @@ function SessionTimeline({
...
   handleActivateBlockById,
+  handleActiveBlock,
+  handleNoActiveBlock,
   subscribeToMore,
 }: Props): React.ReactElement {
   useEffect((): void => {
     subscribeToMore()
   }, [])
 
+  let isBlockActive = activeStep % 2 === 1
```
rschlaefli: Not sure why this is necessary, where do you use `isBlockActive` outside of the `useEffect`? It would be bad practice to alter a variable from within `useEffect`, so if that is needed here you would need to `useState`. But I think you could just remove this as `isBlockActive` is only used inside of `useEffect` and will always be computed on the first render as well.

rschlaefli (changes requested):
Great, just marked a final issue, then LGTM.

rschlaefli (approved):

rschlaefli (approved):

src/components/sessions/SessionTimeline.tsx
lulocher: Alright, makes sense!

PR #1745, opened 2020-10-20
Admin session management

lulocher:
UI for the session management functionality:

- Includes a new tab in the admin area. It displays a table with all currently running sessions, which can now be aborted by an admin.
- Still issues with eslint.


src/components/admin/RunningSessionList.tsx
```diff
@@ -0,0 +1,131 @@
+import React, { useState } from 'react'
+import dayjs from 'dayjs'
+import { useMutation, useQuery } from '@apollo/react-hooks'
```
rschlaefli: Use @apollo/client

PR #2553, opened 2021-10-08
enhance: add subscriptions for unpublishing feedback and deletion of feedback responses

sjschlapbach:
Added new subscription Queries for the deletion of feedbacks and the unpublishing of features using moderator features


frontend/src/components/sessions/join/FeedbackArea.tsx
```diff
@@ -114,11 +128,36 @@ function FeedbackArea({
       },
     })
 
+    const feedbackResponseDeleted = subscribeToMore({
+      document: FeedbackResponseDeletedSubscription,
+      variables: { sessionId },
+      updateQuery: (prev, { subscriptionData }) => {
+        if (!subscriptionData.data) return prev
+        console.log(prev, subscriptionData.data.feedbackResponseDeleted)
```
rschlaefli: remove unnecessary console logging

frontend/src/pages/questions/index.tsx
```diff
@@ -143,7 +143,19 @@ function Index(): React.ReactElement {
...
+        //refetchQueries: [{ query: QuestionPoolQuery }],
+        update: (store, { data: { selectedItems } }) => {
+          const data = store.readQuery({
+            query: QuestionPoolQuery,
+          })
+          // @ts-ignore
+          data.questions = data.questions.map((question, index) =>
+            question.id === selectedItems[index].id
+              ? (question.isArchived = true)
+              : (question.isArchived = question.isArchived)
```
rschlaefli: what does this do? not sure what the entire update function is used for

rschlaefli: if it does something useful, refactor the code so it does not explicitly assign to the question parameter (does not correspond to the principle of mapping => return the new and updated question value, don't assign to a property of it)

rschlaefli (changes requested):
@sjschlapbach Thanks a lot, just some minor changes before merging :)

frontend/src/components/sessions/join/FeedbackArea.tsx
sjschlapbach: missed that, is removed now

frontend/src/pages/questions/index.tsx
sjschlapbach: refactored it; function updates the apollo cache so that hopefully no update through the database is required anymore afterwards

PR #2619, opened 2021-11-05
feat(frontend): reintroduce confusion barometer

sjschlapbach:
This implementation reintroduces a reworked and simplified version of the confusion barometer. It also includes various improvements and refactorings on the overall layout of the application.


frontend/src/pages/join/[shortname].tsx
```diff
@@ -122,25 +122,32 @@ function Join({ shortname }): React.ReactElement {
...
+          variables: {
+            speed,
+            difficulty,
+            fp: fingerprint,
+            sessionId,
+          },
+        })
+        window.sessionStorage?.setItem(
+          `${shortname}-${sessionId}-confusion`,
+          JSON.stringify({
```
Comment (rschlaefli) [inline frontend/src/pages/join/[shortname].tsx]: maybe we should use `localForage` as an abstraction (and thus local storage) and can define our own expiration. that way, we don't have to deal with stringify etc., and the buttons persist across browser tabs. because we use the sessionId in the key, the buttons will not persist across sessions.

frontend/src/pages/join/[shortname].tsx
```diff
@@ -322,9 +332,10 @@ function Join({ shortname }): React.ReactElement {
 
         .joinSession {
           display: flex;
-          min-height: -moz-calc(100vh - 8rem);
-          min-height: -webkit-calc(100vh - 8rem);
-          min-height: calc(100vh - 8rem);
+          height: -moz-calc(100vh - 8rem);
```
Comment (rschlaefli) [inline frontend/src/pages/join/[shortname].tsx]: we should remove the manual `-moz-calc` etc. here, as `autoprefixer` in our css pipeline should automatically create these browser-specific classes.

rschlaefli (changes requested):

frontend/package.json
```diff
@@ -96,6 +96,7 @@
     "react-dropzone": "11.4.2",
     "react-focus-lock": "2.5.2",
     "react-ga": "3.3.0",
+    "react-gauge-chart": "^0.4.0",
```
rschlaefli: also, the `react-gauge-chart` depencency breaks the build `npm run build` on my machine. it probably does not play well with `typescript` yet (e.g., has no types). this can happen fairly often, so always make sure the build still runs through after installing new dependencies.

backend/src/services/sessionExec.js
```diff
@@ -360,11 +361,11 @@ const addConfusionTS = async ({ sessionId, difficulty, speed }) => {
   // extract the saved confusion timestep and convert it to a plain object
   // then readd the mongo _id field under the id key and publish the result
   // this is needed as redis swallows the _id field and the client could break!
-  // const savedConfusion = session.confusionTS[session.confusionTS.length - 1].toObject()
-  // pubsub.publish(CONFUSION_ADDED, {
-  //   [CONFUSION_ADDED]: { ...savedConfusion, id: savedConfusion._id },
-  //   sessionId,
-  // })
+  const savedConfusion = session.confusionTS[session.confusionTS.length - 1].toObject()
```
rschlaefli, line 390: this subscription is not currently used. here, the backend pushes a new confusionTS as incoming directly to the subscription on the client. however, as we have commented out `subscribeToMore` on the client, the subscription is never created, and these events disappear in nirvana.

at some point, we should probably reintroduce a working subscription for the confusion barometer, so the lecturer gets a nice real-time view on the gauge.

frontend/lang/en.json
```diff
@@ -221,7 +223,7 @@
   "runningSession.checkboxes.unpublished": "Unpublished",
   "runningSession.confusion.difficulty.Range": "easy - hard",
   "runningSession.confusion.difficulty.Title": "Difficulty",
-  "runningSession.confusion.info": "The Confusion-Barometer allows you to get feedback on the speed and difficulty of your lecture as it evolves over time.",
+  "runningSession.confusion.info": "The Confusion-Barometer allows you to get feedback on the speed and difficulty of your lecture as it evolves over time. It displays all feedbacks that were shared by students throughout the past 10 minutes",
```
rschlaefli: when I run `npm run build` on my machine, it extracts all of the language strings and builds this `en.json` again. this string here is removed entirely, as it is not used anymore in the code. if we quickly run `npm run build` when a branch is ready to merge, we can ensure that no old strings are kept around in the code. to remove them from `de.json` as well, we would need to remove them by hand and run `npm run build` again, so they are also removed from the compiled version of `de`

for this string here, we could maybe keep it around and put it inside a tooltip. maybe you could investigate adding a tooltip in the confusion barometer view

frontend/src/pages/sessions/running.tsx
```diff
@@ -269,6 +271,7 @@ function Running(): React.ReactElement {
               sessionId={id}
               sessionName={name}
               subscribeToMore={subscribeToMore}
+              confusionFlag={featureFlags?.flags?.confusionBarometer}
```
rschlaefli: on boolean props like this, I tend to use naming like "isX" or "withX", "hasX", so "hasConfusionFlag"

frontend/src/pages/join/[shortname].tsx
```diff
@@ -122,25 +122,32 @@ function Join({ shortname }): React.ReactElement {
...
+      try {
+        newConfusionTS({
+          variables: {
+            speed,
+            difficulty,
+            fp: fingerprint,
+            sessionId,
+          },
+        })
+        window.sessionStorage?.setItem(
```
Comment (rschlaefli) [inline frontend/src/pages/join/[shortname].tsx]: not sure whether it makes sense that the saving to storage is also debounced, or whether that should better happen immediately (e.g., if you change the slider and it does not save immediately, you could have an old state when refreshing)

frontend/src/components/sessions/join/FeedbackArea.tsx
```diff
@@ -205,6 +215,35 @@ function FeedbackArea({
...
+      console.error(e)
+    }
+  }, [])
+
+  const onNewConfusionTS = async (newValue: any, selector: string) => {
+    // send the new confusion entry to the server
+    if (selector === 'speed') {
+      setConfusionSpeed(newValue)
+      handleNewConfusionTS({
+        difficulty: confusionDifficulty ? confusionDifficulty : 0,
```
rschlaefli: this is mosly the same as `confusionDifficulty || 0` or if we want to have a default case for `undefined` -> `confusionDifficulty ?? 0`

frontend/src/components/sessions/join/FeedbackArea.tsx
```diff
@@ -271,7 +310,7 @@ function FeedbackArea({
   return (
     <div
       className={clsx(
-        'bg-white p-2 md:p-4 flex-col md:border-primary md:border-solid md:border flex-1 md:flex',
+        'bg-white p-2 md:p-4 flex-col md:border-primary md:border-solid md:border flex-1 md:flex h-[93vh] md:h-full',
```
rschlaefli: using `vh` for styling can cause a lot of problems. I would restrict the `vh` styles to the platform they are useful for. some browsers (e.g., on mobile), compute `vh` differently (e.g., I believe the browser title bar is included vs. excluded in the total view height on some browsers).

frontend/src/components/sessions/join/FeedbackArea.tsx
```diff
@@ -205,6 +215,35 @@ function FeedbackArea({
     }))
   }, [data?.joinQA, upvotedFeedbacks, reactions])
 
+  useEffect((): void => {
+    try {
+      const confusion = window.sessionStorage?.getItem(`${shortname}-${sessionId}-confusion`)
```
rschlaefli: in my local instance, this does not seem to repopulate the buttons when I reload the tab. will investigate further.

frontend/src/components/interaction/confusion/ConfusionSection.tsx
```diff
@@ -1,76 +1,20 @@
...
-                <Legend />
-              </LineChart>
-            </ResponsiveContainer>
-          )
-        })()}
+    <div>
+      <h3 className="inline-block mr-3">{title}</h3>
+      <p className="inline-block">{'(' + xlabel + ')'}</p>
+      <div className="w-full h-38">
+        <GaugeChart id="gauge-chart" nrOfLevels={20} percent={runningValue} hideText={true} animate={false} />
```
rschlaefli: the `gauge-chart` docs mention not reusing the same `id` across charts. not sure whether that is a factor here, but maybe we need to make the `id` a prop if it causes problems. in any case, `id` should never be used twice in a page (and in this case, it will be, as `ConfusionSection` is a component that is shown multiple times).

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      <Button
+        onClick={(): void => {
+          handleChange(props.onChangeValue)
+          setButtonState(false)
+          clearTimeout()
+          setTimeout(setButtonState, 60000, true)
+        }}
+        className="min-w-[32%] md:min-w-[0%]"
+        color={value == props.onChangeValue ? 'blue' : null}
```
rschlaefli: always use `===` for comparisons (eslint should mention that as well). for the default case of this, I would set it to `undefined`, as we don't define a color if the prop value is not what we check for

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+        disabled={!buttonState}
+      >
+        {props.children}
+      </Button>
+    )
+  }
+
+  return (
+    <div className="mb-10 confusionSlider">
+      <Head>{createLinks(['https://unpkg.com/react-rangeslider/umd/rangeslider.min.css'])}</Head>
```
rschlaefli: this is probably old code left-over from the rangeslider I had used before. will remove, so we save the loading of this css on the join page

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [buttonState, setButtonState] = useState(true)
+
+  const ConfusionButton = (props) => {
+    return (
+      <Button
+        onClick={(): void => {
+          handleChange(props.onChangeValue)
+          setButtonState(false)
+          clearTimeout()
+          setTimeout(setButtonState, 60000, true)
```
rschlaefli: not sure about this setTimeout logic, need to check something

frontend/src/components/interaction/confusion/ConfusionCharts.tsx
```diff
@@ -26,51 +25,45 @@ const messages = defineMessages({
...
+  const [speedRunning, setSpeedRunning] = useState(0)
+  const [difficultyRunning, setDifficultyRunning] = useState(0)
 
-  const parsedTS = confusionTS.reduce((acc, { createdAt, speed, difficulty }): any[] => {
-    const tempAcc = [...acc, { difficulty, speed }]
+  useEffect(() => {
+    const filteredConfusion = confusionTS.filter(
+      (element: any) => dayjs().diff(dayjs(element.createdAt), 'minute') <= 10
+    )
+    const reducerSpeed = (previousValue, currentValue) => previousValue + (currentValue.speed + 1) * 0.5
```
rschlaefli: these computations should probably be moved to the server as well (e.g., so we can have an average over the entire session, but don't have to send all of the data to the client). we could optimize further and compute the "running average" instead if recomputing, by storing the necessary values (e.g., the sum and count of the previous computation, and simply adding to it).

frontend/src/components/interaction/confusion/ConfusionBarometer.tsx
```diff
@@ -16,85 +15,29 @@ interface Props {
...
-          }
-        }
-      `}</style>
-    </div>
-  )
+function ConfusionBarometer({ confusionTS, subscribeToMore }: Props): React.ReactElement {
+  const [forceRerender, setForceRerender] = useState(0)
+
+  useEffect(() => {
+    const timeoutHandle = window.setTimeout(() => {
```
rschlaefli: this should also have a cleanup function so the timeout is purged.

we should call `clearTimeout` in the cleanup function we return in this `useEffect`.

also, without any dependencies, this `useEffect` will run on every render. and will probably be recreated on every render with an increment of forceRerender. this probably works, but causes much more load than running it once every 60 seconds as probably intended.

maybe we would need to add a dependency array and add forceRerender as a dep there, so if the timeout increments the number, the useEffect will run again and create the next timeout

frontend/package.json
rschlaefli: we want dependency versioning to be pinned, e.g., without ^. to achieve this, always install packages with `npm install --save-exact XYZ`. or remove this ^ after install and rerun `npm install` to regenerate package-lock.json

frontend/src/pages/sessions/running.tsx
sjschlapbach: changed with the next commit

frontend/src/pages/join/[shortname].tsx
Comment (sjschlapbach) [reply frontend/src/pages/join/[shortname].tsx]: tried that before, but there were some issues with data that was overwritten in a strange fashion as one would have to use state for the other of the two variables. I could give it another try, but this will required handling of more different cases - e.g. only one of the attributes was set before the reload, etc.
Only option to go around these issues: specify two separate storage items for the two values

frontend/src/components/sessions/join/FeedbackArea.tsx
sjschlapbach: will be changed with the next commit

frontend/src/components/interaction/confusion/ConfusionSection.tsx
sjschlapbach: apparently missed this when programming the component - the id now contains the title of the chart, which shouldn't be identical for both charts so an additional prop was not required

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
sjschlapbach: will be changed with the next commit

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
sjschlapbach: will be done with next commit

frontend/src/components/interaction/confusion/ConfusionBarometer.tsx
sjschlapbach: cleanup function is added in the next commit; probably we should first check if an implementation of the running average on server-side works as expected, as we can probably omit all of this hacky rerender-stuff in a next step

frontend/package.json
sjschlapbach: The '^' will be removed with the next commit (with adapted package-lock file). If this however breaks the build, we will probably have to look for an alternative to illustrate the confusion stuff

backend/src/services/sessionExec.js
sjschlapbach: probably we should move the average computation to the backend ahead of release anyways. Maybe we can then combine this step with reintroducing the subscribeToMore

frontend/lang/en.json
sjschlapbach: The tooltip is a great idea, will definitely look into it - so we can omit all disruptive text in this view

frontend/src/pages/join/[shortname].tsx
Comment (sjschlapbach) [reply frontend/src/pages/join/[shortname].tsx]: removed browser specific stuff - works in all tested browsers

frontend/src/pages/join/[shortname].tsx
Comment (sjschlapbach) [reply frontend/src/pages/join/[shortname].tsx]: also change to localForage

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
sjschlapbach: rename to isButtonDisabled or similar, check implementation clearTimeout (use handle)

frontend/src/components/interaction/confusion/ConfusionCharts.tsx
sjschlapbach: have been outsourced to the backend

frontend/src/components/interaction/confusion/ConfusionBarometer.tsx
sjschlapbach: force rerendering removed; implementation of computation in backend solves this issue

frontend/src/components/interaction/confusion/ConfusionDialog.tsx
sjschlapbach: done - logic tested

frontend/src/components/sessions/join/FeedbackArea.tsx
sjschlapbach: seems to work currently - will be reworked anyways when redesigning the layout for mobile devices

frontend/lang/en.json
sjschlapbach: tooltip was added

PR #2637, opened 2021-11-25
enhance(frontend): improve layout of running session page

sjschlapbach:
The running session page on the lecturer side  is still built using style jsx. This is to be replaced by a more flexible tailwind css implementation. Furthermore, the mobile view of this page is suboptimal at the moment and is to be enhanced.


rschlaefli (approved):
LGTM

PR #2650, opened 2021-12-05
enhance(backend): populate new user accounts with demo data

sjschlapbach:
The accounts of newly registrated users are automatically populated with some demo data including demo questions for every type, a demo session, and an executed demo session with responses, feedbacks, confusion, etc.


apps/backend/src/services/sessionMgr.js
```diff
@@ -941,6 +941,8 @@ async function activateBlockById({ userId, sessionId, blockId }) {
   // find the next block for the running session
   const newBlock = session.blocks[blockIndex]
 
+  console.warn(session.blocks, blockId, blockIndex, newBlock)
```
rschlaefli: removed in next commit

apps/backend/src/services/accounts.js
```diff
@@ -345,155 +387,64 @@ async function hydrateDemoData({ userId }) {
...
-  const demoSession = new SessionModel({
-    settings: {
-      isParticipantAuthenticationEnabled: false,
-      isConfusionBarometerActive: false,
-      isEvaluationPublic: false,
-      isFeedbackChannelActive: false,
-      isFeedbackChannelPublic: true,
-      authenticationMode: 'NONE',
-      storageMode: 'SECRET',
+  const feedbacks = [
```
rschlaefli, line 390: we could also add feedbacks to the "CREATED" session, and they would show up immediately on the running session cockpit after starting it (so there is some data there). confusion values would work as well, but would be filtered out of the chart

apps/backend/src/services/accounts.js
```diff
@@ -176,137 +172,141 @@ async function hydrateDemoData({ userId }) {
...
-    return instance.id
-  }
 
-  // instance array containing instances for questions [1, 1, 2, 2, 3, 3, 4, 4, 1, 2, 3, 4]
-  for (let i = 0; i < titles.length; i += 1) {
-    instanceIds.push(await questionInstanceCreator(demoQuestions[i], sessionId1))
-    instanceIds.push(await questionInstanceCreator(demoQuestions[i], sessionId1))
-  }
+  // prepare 8 question instances for the session without results (in pairs of two)
+  const questionInstances = await demoQuestions.reduce(async (acc, question) => {
```
rschlaefli, line 276: I adapted it like this for the two sessions for simplicity. The order is not consistent across users (q4 might be a SC for one, and NR for another), but here the pairs are consistent (always SC, SC, NR, NR, etc.). The inconsistency actually does not matter, so it was much easier to do everything "in parallel" instead of sequentially with for loops.

apps/backend/src/services/accounts.js
```diff
@@ -176,137 +172,141 @@ async function hydrateDemoData({ userId }) {
...
-  const demoQuestions = []
+  },
+]
+
+function prepareDemoSessionData({ id, name, user, blockInstances, blockStatus = 'PLANNED', ...rest }) {
+  return new SessionModel({
+    _id: id,
+    settings: {
+      isParticipantAuthenticationEnabled: false,
+      isConfusionBarometerActive: true,
```
rschlaefli, line 207: Changed the default here so the confusion barometer is active immediately on the started demo session

apps/backend/src/services/accounts.js
```diff
@@ -176,137 +172,141 @@ async function hydrateDemoData({ userId }) {
...
         restrictions: { min: 0, max: 8000 },
-        choices: [],
       },
     },
-  ]
-  const demoQuestions = []
+  },
+]
+
+function prepareDemoSessionData({ id, name, user, blockInstances, blockStatus = 'PLANNED', ...rest }) {
```
rschlaefli, line 194: ...rest allows us to catch any arguments that are not specifically named, so we can override arbitrary values of the object by using ...rest in the final line below

apps/backend/src/services/accounts.js
```diff
@@ -136,23 +136,14 @@ const checkAvailability = async ({ email, shortname }) => {
...
-  const types = ['SC', 'MC', 'FREE', 'FREE_RANGE']
-  const content = [
-    'Which of the following statements is applicable to _KlickerUZH_?',
-    'Which of the following formulas have the form of a Taylor polynomial of some degree $$n$$: $$T_n f(x;a)$$? (multiple answers are possible)',
-    'Describe a main principle of a social market economy.',
-    'Estimate the length of the **longest** river in the world (answer in kilometres).',
-  ]
-  const options = [
-    {
+const DEMO_QUESTIONS = [
```
rschlaefli, line 139: Adapted this to an array of objects so we can map through without using index based lookups

PR #2723, opened 2022-04-27
fix(backend): double escape backslashes to ensure correct latex rendering

sjschlapbach:
Currently, there are issues with formulas that use special chars that are escaped with backslashes. Our editor library removes those backslashes, what results in problems.


apps/backend/src/services/questions.js
```diff
@@ -304,7 +304,7 @@ const modifyQuestion = async (questionId, userId, { title, tags, content, option
     // push a new version into the question model
     try {
       question.versions.push({
-        content: content || lastVersion.content,
+        content: content.replace(/\\/g, '\\\\') || lastVersion.content.replace(/\\/g, '\\\\'),
```
rschlaefli:
```suggestion
        content: content?.replace(/\\/g, '\\\\') || lastVersion.content?.replace(/\\/g, '\\\\'),
```

PR #2724, opened 2022-05-09
fix: hotfix to add results back to the db that were lost during the last migration

sjschlapbach:
During the last database migration a major part of the past sessions was migrated without results. This should be fixed as soon as possible using the db dump from before the migration.


rschlaefli:
Will be done as part of v2 to v3 migration

PR #2786, opened 2022-08-24
feat: add participant login in PWA

florinavogel:


apps/frontend-pwa/next.config.js
```diff
@@ -1,14 +1,16 @@
...
-const withPWA = require('next-pwa')
+const withPWA = require('next-pwa')({
+  dest: 'public'
+})
 
 const nextConfig = withPWA({
   reactStrictMode: true,
   swcMinify: true,
   pwa: {
     dest: 'public',
```
rschlaefli: This could all be moved to withPWA above

apps/frontend-pwa/package.json
```diff
@@ -15,13 +15,15 @@
     "@klicker-uzh/graphql": "^2.0.0-alpha.8",
     "@uzh-bf/design-system": "0.0.25",
     "deepmerge": "4.2.2",
+    "formik": "^2.2.9",
```
rschlaefli: Always add dependencies using `npm i -E` -> exact specification instead of with a `^`. Can explain later why that is especially important in the new architecture.

apps/frontend-pwa/src/pages/Login.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { ErrorMessage, Field, Form, Formik } from 'formik'
+import { twMerge } from 'tailwind-merge'
+import * as Yup from 'yup'
+
+const loginSchema = Yup.object().shape({
+  username: Yup.string().required('Enter your username'),
+  password: Yup.string().required('Enter your password'),
+})
+
+const LoginForm = () => {
```
rschlaefli:
```suggestion
function LoginForm() {
```

apps/frontend-pwa/src/pages/Login.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const onSubmit = async (values: any, { setSubmitting, resetForm }: any) => {
+    // await loginParticipant(values.username, values.password)
```
rschlaefli: To get this to work: add `useMutation` as a top-level hook, import the LoginParticipantDocument from the prisma package, and pass it to the useMutation hook. Then, await loginParticipant({ variables: { username, password }})

apps/frontend-pwa/src/pages/Login.tsx
```diff
@@ -0,0 +1,96 @@
+import { faArrowRightToBracket } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import {} from '@klicker-uzh/graphql/dist/ops'
+import * as RadixLabel from '@radix-ui/react-label'
+import { Button, H1 } from '@uzh-bf/design-system'
+import { ErrorMessage, Field, Form, Formik } from 'formik'
+import { twMerge } from 'tailwind-merge'
+import * as Yup from 'yup'
+
+const loginSchema = Yup.object().shape({
```
rschlaefli: Yup is great for validation with formik :) Maybe we could add some requirements here (username > 4 characters, etc.). Also to be validated in the backend, of course.

apps/frontend-pwa/src/pages/Login.tsx
```diff
@@ -0,0 +1,96 @@
...
+                htmlFor="username"
+                className="text-sm leading-7 text-gray-600"
+              >
+                Username
+              </RadixLabel.Root>
+              <Field
+                name="username"
+                type="text"
+                className={twMerge(
+                  'w-full px-3 py-1 text-base leading-8 text-gray-700 transition-colors',
```
rschlaefli: We can extract this to a generic Input component later

packages/graphql/src/graphql/schema.graphql
```diff
@@ -28,7 +28,8 @@ type LearningElement {
 type Mutation {
   joinCourse(courseId: ID!): Participation
   leaveCourse(courseId: ID!): Participation
-  login(email: String!, password: String!): ID
+  loginParticipant(password: String!, username: String!): ID
```
rschlaefli, line 57: Preferably use the same ordering as the other mutation (username, pw)

packages/graphql/src/services/accounts.ts
```diff
@@ -1,15 +1,24 @@
...
 import { Context } from '../lib/context'
 
-interface LoginArgs {
+interface LoginUserArgs {
   email: string
   password: string
 }
 
-export async function login({ email, password }: LoginArgs, ctx: Context) {
+interface LoginParticipantArgs {
```
rschlaefli: Place this above the corresponding function (i.e., line 64)

packages/graphql/src/services/accounts.ts
```diff
@@ -53,3 +62,47 @@ export async function login({ email, password }: LoginArgs, ctx: Context) {
...
+  const jwt = JWT.sign(
+    {
+      sub: participant.id,
+      role: UserRole.PARTICIPANT,
+    },
+    // TODO: use structured configuration approach
+    process.env.APP_SECRET as string,
+    {
+      algorithm: 'HS256',
+      expiresIn: '6d',
```
rschlaefli: 7d here, 6d in the cookie -> the cookie should expire before the token

rschlaefli: But to be adapted to "refresh tokens" anyways

packages/graphql/src/services/accounts.ts
```diff
@@ -53,3 +62,47 @@ export async function login({ email, password }: LoginArgs, ctx: Context) {
...
+    }
+  )
+
+  ctx.res.cookie('user_token', jwt, {
```
rschlaefli:
```suggestion
  ctx.res.cookie('participant_token', jwt, {
```

`user_token` is reserved for normal users and admins -> we have two so that admin users can also impersonate a participant without being logged out from the admin view

packages/graphql/src/services/accounts.ts
```diff
@@ -53,3 +62,47 @@ export async function login({ email, password }: LoginArgs, ctx: Context) {
...
+    domain: process.env.COOKIE_DOMAIN ?? process.env.API_DOMAIN,
+    path: '/',
+    httpOnly: true,
+    maxAge: 60 * 60 * 24 * 6,
+    secure: process.env.NODE_ENV === 'production',
+  })
+
+  return participant.id
```
rschlaefli, line 108: Might be sensible to return some more data on the participant here (e.g., for immediate display in the app). For example, the avatar, etc. -> Can discuss this later

packages/prisma/src/prisma/schema.prisma
```diff
@@ -338,10 +338,11 @@ model ParticipantAccount {
 model Participant {
   id String @id @default(uuid()) @db.Uuid
 
-  pseudonym String @unique
-  email     String @unique
-  password  String
-  avatar    String @default("placeholder.png")
+  pseudonym   String    @unique
```
rschlaefli:
```suggestion
  username   String    @unique
```

rschlaefli (changes requested):

apps/frontend-pwa/src/pages/Login.tsx
florinavogel: I had something like this in the code initially, but then I thought it makes more sense in a register page than in a login page where users have already chosen their username.

packages/prisma/src/prisma/schema.prisma
florinavogel: Let's hope I changed it in all the necessary places :)

PR #2800, opened 2022-08-27
feat: rebuild join page

sjschlapbach:
Rebuild the entire join page for Klicker v2 including all required database queries and backend services


apps/frontend-pwa/src/components/questions/NUMERICALAnswerOptions.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function NUMERICALAnswerOptions({
+  valid,
+  value,
+  onChange,
+}: NUMERICALAnswerOptionsProps): React.ReactElement {
+  return (
+    <div className="flex flex-col gap-2">
+      <textarea
```
rschlaefli: For numbers, much more useful to use an input with type="number"

apps/frontend-pwa/src/constants.ts
```diff
@@ -0,0 +1,201 @@
...
+  rgb(217, 78, 31),  // 6
+  rgb(241, 108, 32), // 7
+  rgb(239, 139, 44), // 8
+  rgb(236, 170, 56), // 9
+  rgb(235, 200, 68), // 10
+  rgb(162, 184, 108),// 11
+  rgb(92, 167, 147),// 12
+] */
+
+// export const CHANGELOG = {
```
rschlaefli: all of this old bloat can be removed

package.json
```diff
@@ -54,5 +54,8 @@
   "packageManager": "npm@8.15.0",
   "engines": {
     "node": "=18"
+  },
+  "dependencies": {
+    "turbo-darwin-64": "^1.4.3"
```
rschlaefli: don't add this to deps, the install of this might fail on linux or windows. just install it globally

rschlaefli: there should be a better solution to this problem, but for now, it works

packages/graphql/src/services/sessions.ts
```diff
@@ -54,7 +54,11 @@ export async function getSession({ id }: { id: string }, ctx: ContextWithUser) {
   const session = await ctx.prisma.session.findUnique({
     where: { id },
     include: {
-      blocks: true,
+      blocks: {
+        include: {
+          instances: true,
```
rschlaefli, line 337: parse out the solutions of in question data on the server side

apps/backend-sls/prisma/seed.ts
```diff
@@ -282,6 +282,50 @@ async function main(prisma: Prisma.PrismaClient) {
...
+    create: {
+      id: '996f208b-d567-4f1e-8c57-6f555866c33f',
+      name: 'RDWachstum',
+      content:
+        'Wie hoch schätzt du das durchschnittliche Wachstum der Forschung- und Entwicklungskosten in den nächsten fünf Jahren?',
+      contentPlain:
+        'Wie hoch schätzt du das durchschnittliche Wachstum der Forschung- und Entwicklungskosten in den nächsten fünf Jahren?',
+      type: 'NUMERICAL',
+      options: {
+        restrictions: {
```
rschlaefli: the session management branch adds seeding for numerical and free text as well (will conflict). add solutions to the seed

apps/frontend-pwa/package.json
```diff
@@ -17,18 +17,31 @@
     "@fortawesome/free-solid-svg-icons": "6.1.2",
     "@fortawesome/react-fontawesome": "0.2.0",
     "@klicker-uzh/graphql": "^2.0.0-alpha.8",
-    "@uzh-bf/design-system": "0.0.26",
+    "@radix-ui/react-progress": "^1.0.0",
```
rschlaefli: remove the ^ in deps

rschlaefli: also, progress should be part of the design system already?

apps/frontend-pwa/package.json
```diff
@@ -17,18 +17,31 @@
...
+    "dayjs": "1.11.5",
     "deepmerge": "4.2.2",
     "formik": "2.2.9",
     "graphql": "16.6.0",
+    "localforage": "^1.10.0",
     "next": "12.2.5",
     "next-pwa": "5.6.0",
     "ramda": "0.28.0",
     "react": "18.2.0",
+    "react-countdown-circle-timer": "3.0.9",
```
rschlaefli: might also move this to the design system

apps/frontend-pwa/package.json
```diff
@@ -17,18 +17,31 @@
...
     "react-dom": "18.2.0",
     "react-loader-spinner": "5.3.3",
+    "rehype-katex": "6.0.2",
```
rschlaefli: design system component or separate package for markdown logic?

rschlaefli: I think this logic for markdown parsing might be useful in several of our apps

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
```diff
@@ -0,0 +1,311 @@
...
+    inputEmpty: true,
+    inputValid: false,
+    inputValue: QUESTION_GROUPS.CHOICES.includes(
+      questions[remainingQuestions[0]]?.type
+    )
+      ? new Array(questions[remainingQuestions[0]]?.options.length, false)
+      : undefined,
+  })
+
+  // TODO: fix response storage, currently multiple replays are possible - maybe automatically resolved with handleNewResponse implementation
```
rschlaefli, line 60: this will only be needed when responding anonymously. if logged in, we can verify server-side that a participant does not respond twice.

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
```diff
@@ -0,0 +1,311 @@
+import { push } from '@socialgouv/matomo-next'
+import { H1 } from '@uzh-bf/design-system'
+import dayjs from 'dayjs'
+import localForage from 'localforage'
+import _without from 'lodash/without'
```
rschlaefli: move from lodash stuff to only using ramda

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
```diff
@@ -0,0 +1,311 @@
...
+import React, { useEffect, useState } from 'react'
+import v8n from 'v8n'
```
rschlaefli: I think we use several different validation libraries across the code base. should try to find a common library that works everywhere. validator, v8n, yup, etc.

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+          <div className="flex-initial min-h-[6rem] p-3 bg-primary-10 border-uzh-blue-80 border border-solid rounded">
+            <QuestionDescription
+              content={currentQuestion.content}
+              description={currentQuestion.description}
+            />
+          </div>
+
+          {/* // TODO */}
+          <div>QUESTIONFILES</div>
```
rschlaefli, line 291: display attachments from the database. use next/image

apps/frontend-pwa/src/pages/session/[id].tsx
```diff
@@ -1,27 +1,95 @@
...
+
+import FeedbackArea from '@components/liveSession/FeedbackArea'
+import QuestionArea from '@components/liveSession/QuestionArea'
+import { initializeApollo } from '@lib/apollo'
+import Head from 'next/head'
 import { useRouter } from 'next/router'
+import { twMerge } from 'tailwind-merge'
 
-function Index() {
+function Index({ session }: any) {
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/session/[id].tsx:20]: use the type that is returned from the query in server side props. otherwise, we lose all of the typings here

apps/frontend-pwa/src/pages/session/[id].tsx
```diff
@@ -1,27 +1,95 @@
...
+        <div className="w-1/2 p-4 bg-white border-2 border-solid rounded-lg border-uzh-blue-40">
+          <FeedbackArea feedbacks={[]} />
+        </div>
+      )}
     </div>
   )
 }
 
+export const getServerSideProps: GetServerSideProps = async (ctx) => {
+  const apolloClient = initializeApollo()
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/session/[id].tsx:124]: use `addApolloState` similar to usage in `frontend-learning/pages/courseid/index`. this should allow to not explicitly pass the session, but to use the same query based on the cache on the client, without refetching. best to try this. should also solve the typing issue with props

apps/frontend-pwa/src/pages/session/[id].tsx
```diff
@@ -1,27 +1,95 @@
...
+
+  // TODO: implement polling, to be replaced by more efficient solution like subscriptions
+  const result = await apolloClient.query({
+    query: GetSessionDocument,
+    variables: {
+      id: ctx.query?.id as string,
+    },
+  })
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/session/[id].tsx:132]: use skip to skip querying if the ctx.query.id is not defined. then assume in variables that it is set. prevents double querying (without an id and with an id again)

apps/frontend-pwa/src/pages/session/[id].tsx
```diff
@@ -1,27 +1,95 @@
...
+
+  return { props: { session: result.data?.session } }
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/session/[id].tsx:134]: set-up incremental static regeneration with fallback, etc.

apps/frontend-pwa/src/components/questions/NUMERICALAnswerOptions.tsx
sjschlapbach: used input text for now as input with type="number" just returns empty strings whenever there are characters, etc. in there, which is unsuitable for form validation on the level above

apps/frontend-pwa/package.json
sjschlapbach: removed unused dependency

apps/frontend-pwa/package.json
sjschlapbach: removed unused dependency

apps/frontend-pwa/src/constants.ts
sjschlapbach: removed stuff that will not be used anyways

package.json
sjschlapbach: removed dependency

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
sjschlapbach: replaced v8n with yup and removed corresponding dependency

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
sjschlapbach: replaced with ramda function

apps/frontend-pwa/src/components/liveSession/QuestionArea.tsx
sjschlapbach: mapping that maps attachments to next/images if they are defined will be implemented

apps/frontend-pwa/src/pages/session/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/session/[id].tsx:132]: This skip option seems to exist primarily on the useQuery hook. TODO: investigate whether it is somehow still possible to use it or implement an alternative with the same functionality

apps/backend-sls/prisma/seed.ts
sjschlapbach: naming schemes will be adapted to the session management branch in order to simplify merging

apps/frontend-pwa/package.json
sjschlapbach: moved to separate package within the klicker-uzh repository

PR #2819, opened 2022-09-02
feat: initial session creation and start 

rschlaefli:


apps/backend-sls/GraphQL/index.ts
```diff
@@ -1,9 +1,33 @@
+import { PrismaClient } from '@klicker-uzh/prisma'
 import serverlessExpress from '@vendia/serverless-express'
+import Redis from 'ioredis'
 
-import app from './app'
+import prepareApp from './app'
+
+const prisma = new PrismaClient()
```
rschlaefli, line 8: Prisma and Redis connections are initiated outside of the main handler that we return -> might be reused across serverless function instances, if we are lucky

packages/prisma/src/prisma/schema.prisma
```diff
@@ -104,7 +104,7 @@ enum QuestionType {
 }
 
 model Question {
-  id String @id @default(uuid()) @db.Uuid
+  id Int @id @default(autoincrement())
```
rschlaefli, line 107: Migrated everything that we do not expose for public access (e.g., a join link) to int ids for efficiency

packages/graphql/src/graphql/schema.graphql
```diff
@@ -144,7 +152,7 @@ enum SessionBlockStatus {
 
 enum SessionStatus {
   COMPLETED
-  PLANNED
+  PREPARED
```
rschlaefli, line 201: A session is now "PREPARED" after the initial creation (as having both "PLANNED" and "SCHEDULED" does not make sense). "SCHEDULED" -> might use this in the future to say that a session will automatically start at a predefined time, for example.

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
+import { PrismaClient } from '@klicker-uzh/prisma'
```
rschlaefli, line 1: Created an initial integration test that logs in with the seeded user and creates a session, starts it, and activates a block. Interacts with the database and redis cache, though much is still missing.

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
...
 import request from 'supertest'
+import prepareApp from '../GraphQL/app'
 
-import app from '../src/app'
+// TODO: switch to ioredis-mock
+// jest.mock('ioredis', () => Redis)
 
-describe('GraphQL API', () => {
-  it('works', async () => {
+process.env.NODE_ENV = 'development'
```
rschlaefli, line 9: If this is not set, GraphQL Yoga will swallow errors and only display "internal error" or similar for anything that breaks inside a service/resolver, which caused me to debug for a few hours unnecessarily.

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
+import { PrismaClient } from '@klicker-uzh/prisma'
+import Redis from 'ioredis'
 import request from 'supertest'
+import prepareApp from '../GraphQL/app'
 
-import app from '../src/app'
+// TODO: switch to ioredis-mock
+// jest.mock('ioredis', () => Redis)
```
rschlaefli, line 7: This would allow us to simulate redis in-memory without having to run it in the test environment. Did not get it to work yet.

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
...
+    })
+  })
+
+  afterAll(async () => {
+    await redisCache.quit()
+    await redisExec.quit()
+  })
+
+  beforeEach(() => {
+    app = prepareApp({ prisma, redisCache, redisExec })
```
rschlaefli, line 43: Here we reuse the same factory function as in the backend index.ts

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
...
+
+    expect(response.headers['set-cookie']).toBeDefined()
+    userCookie = response.headers['set-cookie'][0]
+
+    expect(response.body).toMatchInlineSnapshot(`
+      Object {
+        "data": Object {
+          "loginUser": "6f45065c-447f-4259-818c-c6f6b477eb48",
+        },
+        "extensions": Object {
```
rschlaefli, line 65: This is provided by the response cache plugin, which will dynamically store query and mutation results in memory, and invalidate them with heuristics. Works quite well, but keep in mind when developing (e.g., check that errors are not caused due to caching).

apps/backend-sls/test/integration.test.ts
```diff
@@ -1,13 +1,241 @@
...
+      Object {
+        "data": Object {
+          "startSession": Object {
+            "activeBlock": -1,
+            "blocks": null,
+            "id": "b9fcdfad-f58d-4a44-8dae-a2a5dfe778f8",
+            "status": "RUNNING",
+          },
+        },
+        "extensions": Object {
```
rschlaefli, line 167: The cache for the session that we queried above was invalidated, because we directly mutated the session (the status is now RUNNING) and returned the changes (what I meant by heuristics above).

packages/graphql/src/services/sessions.ts
```diff
@@ -1,13 +1,132 @@
...
 
-export async function startSession(
-  { id }: { id: string },
+function processQuestionData(question: Question): AllQuestionTypeData {
+  switch (question.type) {
+    case QuestionType.SC:
+    case QuestionType.MC: {
+      return {
+        ...question,
+        options: question.options!.valueOf(),
```
rschlaefli, line 15: ! is a declaration like ? in TS, that tells the compiler the value will always be there (so no "might be undefined" warnings). Use carefully (i.e., never).

packages/graphql/src/services/sessions.ts
```diff
@@ -1,13 +1,132 @@
...
+  { name, displayName, blocks }: CreateSessionArgs,
   ctx: ContextWithUser
 ) {
-  const redis = getRedis()
+  const allQuestionsIds = blocks.reduce<number[]>(
+    (acc, block) => [...acc, ...block.questionIds],
+    []
+  )
+
+  const questions = await ctx.prisma.question.findMany({
```
rschlaefli, line 75: `findMany` is a batch query (gets all of the data in a single SQL statement) -> might be more efficient than looping through allQuestionIds above and getting every single question with `findUnique`. Have a look at https://www.prisma.io/docs/guides/performance-and-optimization/query-optimization-performance .

packages/graphql/src/services/sessions.ts
```diff
@@ -1,13 +1,132 @@
...
+      id: {
+        in: allQuestionsIds,
+      },
+    },
+    include: {
+      attachments: true,
+    },
+  })
+
+  const data = {
```
rschlaefli: Looks a bit complex, but is, in principle, just a db "transaction". Creates a session alongside all of the required session blocks and question instances inside of these blocks. Does it all at once, or not at all. And will automatically create the relationships from "Session - SessionBlock - QuestionInstance".

packages/graphql/src/services/sessions.ts
```diff
@@ -23,6 +142,20 @@ export async function startSession(
     return null
   }
 
+  try {
+    const results = await ctx.redisExec
```
rschlaefli: redixExec -> the redis in-memory cache that will store all of the results of a session while it is running (so no DB persistence during a live session, everything is done in-memory). Everything is stored under a "key" in redis, we use something like "s:SESSION\_ID:i:INSTANCE\_ID:results" to store the incoming responses on a specific question, for example.

packages/graphql/src/services/sessions.ts
```diff
@@ -23,6 +142,20 @@ export async function startSession(
...
+      .multi()
+      .hmset(`s:${session.id}:meta`, {
+        id: session.id,
+        namespace: session.namespace,
+        execution: session.execution,
+      })
+      .hset(`s:${session.id}:lb`, { participants: 0 })
```
rschlaefli: `lb` will contain the leaderboard of a live session.

packages/graphql/src/services/sessions.ts
```diff
@@ -50,6 +183,116 @@ export async function startSession(
...
+  ctx: ContextWithUser
+) {
+  const session = await ctx.prisma.session.findUnique({
+    where: {
+      id: sessionId,
+    },
+    include: { blocks: true },
+  })
+
+  if (!session || session.ownerId !== ctx.user.sub) return null
```
rschlaefli, line 216: Because of the efficiency of `findUnique`, this approach to verify ownership should be better than using `findFirst` and including the ownerId as a WHERE clause in the SQL statement. findFirst might cause a full scan of the table, while findUnique should use the index. At least I think so.

packages/graphql/src/services/sessions.ts
```diff
@@ -50,6 +183,116 @@ export async function startSession(
...
+    where: {
+      id: sessionBlockId,
+    },
+    include: {
+      instances: true,
+    },
+  })
+
+  // initialize the cache for the new active block
+  const redisMulti = ctx.redisExec.multi()
```
rschlaefli, line 272: `multi()` together with `exec()` later on are starting a redis transaction. so all of the commands between `multi` and `exec` will be done in one request, and together.

packages/graphql/src/services/sessions.ts
```diff
@@ -50,6 +183,116 @@ export async function startSession(
...
+      case QuestionType.FREE_TEXT: {
+        redisMulti.hmset(`s:${session.id}:i:${instance.id}:results`, {
+          participants: 0,
+        })
+        break
+      }
+    }
+  })
+
+  redisMulti.exec()
```
rschlaefli, line 302: We could `await` this here to get the results of the redis transaction. As I don't care about the results, I don't await and return from the query quicker. But will not get errors thrown if not awaiting the result..

packages/graphql/src/services/sessions.ts
```diff
@@ -50,6 +183,116 @@ export async function startSession(
...
+
+  // TODO: if it has been executed already, rehydrate the cache
```
rschlaefli, line 304: This function still needs a lot of work, but should be appropriate for a first "documentation" of the principles :)

apps/backend-sls/prisma/seed.ts
```diff
@@ -617,6 +619,9 @@ async function main(prisma: Prisma.PrismaClient) {
     },
     update: {},
   })
+
+  await prisma.$executeRaw`ALTER SEQUENCE "Question_id_seq" RESTART WITH 5`
```
rschlaefli: These statements are crucial and need to be kept up-to-date when extending the seed above. Otherwise the integration tests and application will fail, as the automatically created ids will start at 0, but that already exists due to the seeding. Will cause unique index errors.

apps/backend-sls/prisma/seed.ts
```diff
@@ -282,12 +282,62 @@ async function main(prisma: Prisma.PrismaClient) {
...
+      id: 5,
+      name: 'Numerische Testfrage',
+      content: 'Was ist richtig?',
+      contentPlain: 'Was ist richtig?',
+      type: 'NUMERICAL',
+      options: {
+        restrictions: {
+          min: 0,
+        },
+        solutionRanges: [
```
rschlaefli, line 299: Solutions for numerical questions are defined in ranges. A range with only min is exact, meaning that it will only match the exact number. Questions don't need to have a solution (optional), but without solution, they will not be usable for gamified features.

apps/backend-sls/prisma/seed.ts
```diff
@@ -282,12 +282,62 @@ async function main(prisma: Prisma.PrismaClient) {
...
+      id: 6,
+      name: 'Freitext Testfrage',
+      content: 'Was ist richtig?',
+      contentPlain: 'Was ist richtig?',
+      type: Prisma.QuestionType.FREE_TEXT,
+      options: {
+        restrictions: {
+          maxLength: 200,
+        },
+        solutions: ['Schweiz', 'CH'],
```
rschlaefli, line 328: Maybe: define solutions for free text as a list of strings? or no solutions at all?

packages/graphql/src/graphql/ops/FQuestionData.graphql
```diff
@@ -16,5 +16,27 @@ fragment QuestionData on QuestionInstance {
         }
       }
     }
+
+    ... on NumericalQuestionData {
```
rschlaefli, line 20: as the different question types have different schemas for options, we use fragments and spread them based on the parent question type

packages/graphql/src/graphql/schema.graphql
```diff
@@ -35,6 +41,26 @@ A date-time string at UTC, such as 2007-12-03T10:15:30Z, compliant with the `dat
...
+  contentPlain: String!
+  id: Int
+  isArchived: Boolean!
+  isDeleted: Boolean!
+  name: String!
+  options: FreeTextQuestionOptions!
+  type: String!
+}
+
+type FreeTextQuestionOptions {
```
rschlaefli, line 55: These different object types define what exactly is in the options of each question type, and will be used based on the type propertiy of the question (match the corresponding fragment).

packages/graphql/src/nexus.ts
```diff
@@ -49,6 +60,12 @@ export const QuestionData = interfaceType({
     if (item.type === DB.QuestionType.SC || item.type === DB.QuestionType.MC) {
```
rschlaefli, line 60: this logic basically defines, how we choose which type of question options will be returned. we use the type of the question to decide.

packages/graphql/src/nexus.ts
```diff
@@ -35,7 +46,7 @@ export const ResponseInput = inputObjectType({
 export const QuestionData = interfaceType({
```
rschlaefli, line 46: using an interface type here allows us to share all of the common properties of a question, while having different types for the options property.

packages/graphql/src/services/sessions.ts
```diff
@@ -1,13 +1,144 @@
...
+    case QuestionType.MC: {
+      return {
+        ...question,
+        options: question.options!.valueOf(),
+      } as ChoicesQuestionData
+    }
+
+    case QuestionType.NUMERICAL: {
+      return {
+        ...question,
```
rschlaefli, line 21: currently no-op for numerical and free-text, but might be extended

packages/graphql/src/services/sessions.ts
```diff
@@ -1,13 +1,144 @@
...
+    case QuestionType.MC: {
+      const choices = questionData.options.choices.reduce(
+        (acc, _, ix) => ({ ...acc, [ix]: 0 }),
+        {}
+      )
+      return { choices } as ChoicesQuestionResults
+    }
+
+    case QuestionType.NUMERICAL: {
+      return {} as NumericalQuestionResults
```
rschlaefli, line 45: no-op but might be extended

packages/graphql/src/types/app.d.ts
```diff
@@ -52,4 +57,15 @@ interface ChoicesQuestionResults {
...
-type AllQuestionResults = ChoicesQuestionResults
+interface NumericalQuestionResults {
+  [x: string]: number
+}
+
+interface FreeTextQuestionResults {
+  [x: string]: number
+}
+
+type AllQuestionResults =
```
rschlaefli, line 68: these type unions allow TS to infer what kind of data we will have available if we, e.g., use `options` or `results` inside a switch statement which is based on the question.type

packages/prisma/src/prisma/schema.prisma
```diff
@@ -32,7 +32,7 @@ model Account {
   ssoType SSOType
   ssoId   String
 
-  user   User   @relation(fields: [userId], references: [id])
+  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade, onUpdate: Cascade)
```
rschlaefli, line 35: Defining `onDelete: Cascade` will ensure that an entity is deleted if, e.g., the corresponding user account is deleted. otherwise, the operation could fail due to db key constraints. more info see https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions#cascade

packages/prisma/src/prisma/schema.prisma
```diff
@@ -87,9 +87,9 @@ model Attachment {
 
   type AttachmentType
 
-  question   Question? @relation(fields: [questionId], references: [id])
-  questionId String?   @db.Uuid
-  owner      User      @relation(fields: [ownerId], references: [id])
+  question   Question? @relation(fields: [questionId], references: [id], onDelete: SetNull, onUpdate: Cascade)
```
rschlaefli, line 90: Special case: if a question is deleted, we do not want to delete all of the question instances (we want to allow for this use case, hence the copying of questionData into the instance). `SetNull` will ensure that the question property on instances of a deleted question are set to `null`.

packages/prisma/src/prisma/schema.prisma
```diff
@@ -140,23 +140,23 @@ model Tag {
...
   learningElementId String?          @db.Uuid
-  microSession      MicroSession?    @relation(fields: [microSessionId], references: [id])
+  microSession      MicroSession?    @relation(fields: [microSessionId], references: [id], onDelete: Cascade, onUpdate: Cascade)
   microSessionId    String?          @db.Uuid
 
-  question   Question @relation(fields: [questionId], references: [id])
-  questionId String   @db.Uuid
-  owner      User     @relation(fields: [ownerId], references: [id])
-  ownerId    String   @db.Uuid
+  question   Question? @relation(fields: [questionId], references: [id], onDelete: SetNull, onUpdate: Cascade)
```
rschlaefli, line 156: Special case: if a question is deleted, we do not want to delete all of the question instances (we want to allow for this use case, hence the copying of questionData into the instance). \`SetNull\` will ensure that the question property on instances of a deleted question are set to \`null\`.

packages/prisma/src/prisma/schema.prisma
```diff
@@ -239,15 +239,15 @@ model Session {
...
 
   activeBlock Int            @default(-1)
   blocks      SessionBlock[]
 
-  owner   User   @relation(fields: [ownerId], references: [id])
+  owner   User   @relation(fields: [ownerId], references: [id], onDelete: Cascade, onUpdate: Cascade)
   ownerId String @db.Uuid
 
-  course   Course? @relation(fields: [courseId], references: [id])
+  course   Course? @relation(fields: [courseId], references: [id], onDelete: SetNull, onUpdate: Cascade)
```
rschlaefli, line 250: Sessions can exist without courses, so if a course is deleted, its sessions and results could be kept around (but maybe not best user experience).

PR #2851, opened 2022-09-08
feat: block deactivation with redis persistence, leaderboard computation

rschlaefli:


apps/backend-responses/AddResponse/index.ts
```diff
@@ -68,21 +74,10 @@ const httpTrigger: AzureFunction = async function (
 
   const redisMulti = redisExec.pipeline()
 
-  // if we are processing a first response, set the timestamp on the instance
-  // this will allow us to award points for response timing
-  // TODO: do this on the first correct response only
-  if (!firstResponseReceivedAt) {
```
rschlaefli, line 74: moved this so that it is set on the first correct response, not the first ever. the logic below is not DRY yet and can be cleaned up a bit

apps/backend-sls/prisma/constants.ts
```diff
@@ -0,0 +1,12 @@
+export const PARTICIPANT_IDS = [
```
rschlaefli, line 1: participant ids for usage in seed and testing

apps/backend-sls/prisma/seed.ts
```diff
@@ -37,30 +38,34 @@ async function main(prisma: Prisma.PrismaClient) {
...
-    },
-    create: {
-      id: '6f45065c-667f-4259-818c-c6f6b477eb48',
-      email: 'testuser@bf.uzh.ch',
-      password: hash2,
-      username: 'rschlaefli',
-      avatar: 'apple-icon-180.png',
-      participations: {
+  const participants = await Promise.all(
+    PARTICIPANT_IDS.map((id, ix) => {
```
rschlaefli, line 42: creates 10 testusers, all with testuser0 to testuser10 and same password (testing)

apps/frontend-learning/package.json
```diff
@@ -16,7 +16,7 @@
     "@uzh-bf/design-system": "0.0.35",
     "deepmerge": "4.2.2",
     "graphql": "16.6.0",
-    "next": "12.2.5",
+    "next": "12.3.0",
```
rschlaefli, line 19: upgrade in preparation for new features ;)

packages/graphql/src/services/sessions.ts
```diff
@@ -362,17 +360,82 @@ export async function deactivateSessionBlock(
...
-  console.warn(allInstanceResults)
+  if (!cachedResults) return session
 
-  return session
+  const mappedResults: any[] = cachedResults.map(([_, result]) => result)
+
+  const sessionLeaderboard: Record<string, string> = mappedResults[0]
+  const blockLeaderboard: Record<string, string> = mappedResults[1]
+
+  const instanceResults: Record<
```
rschlaefli, line 379: this is critical and a bit complex. the redismulti above returns everything that has been added to the "transaction" in one large array, so e.g., [session leaderboard, session block leaderboard, instance1 response hashes, instance 1 responses, instance 1 results, ...].

all of these items have a different structure, hence cutting off the first two separately, and using a modulo-based approach to select the first, second, and third items of each instance

packages/graphql/src/services/sessions.ts
```diff
@@ -362,17 +360,82 @@ export async function deactivateSessionBlock(
...
+      responses: Record<string, string>
+      results: Record<string, any>
+    }
+  > = mappedResults.slice(2).reduce((acc: any, cacheObj: any, ix) => {
+    const ixMod = ix % 3
+    const instance: QuestionInstance =
+      session.activeBlock!.instances[Math.floor((ix - ixMod) / 3)]
+    switch (ixMod) {
+      // results
+      case 2: {
```
rschlaefli, line 392: would be much nicer to have types defined for the cacheObj depending on the case here, but not implemented right now

packages/graphql/src/services/sessions.ts
```diff
@@ -383,20 +446,93 @@ export async function deactivateSessionBlock(
...
           data: {
             status: SessionBlockStatus.EXECUTED,
             instances: {
-              update: [],
+              update: Object.entries(instanceResults).map(([id, results]) => ({
+                where: { id: Number(id) },
+                data: {
+                  results: results.results,
+                  // TODO: persist responses or "too much information"? delete when session is completed? what about anonymous users?
+                  responses: {
```
rschlaefli, line 459: not sure whether it is any use to persist each response for each participant, or too much data/information. could delete it once the session is completed, or a week after, or so.

packages/graphql/src/services/sessions.ts
```diff
@@ -383,20 +446,93 @@ export async function deactivateSessionBlock(
...
+      },
+    },
+    include: {
+      blocks: true,
     },
   })
 
-  return session
+  // unlink everything regarding the block in redis
+  const unlinkMulti = ctx.redisExec.pipeline()
```
rschlaefli, line 525: unlink deletes the irrelevant stuff in the redis cache (as it has been persisted to the db above). but it does so asynchronously, in a separate thread on redis (i.e., if the cache is not under load, it deletes it)

packages/graphql/src/services/sessions.ts
```diff
@@ -383,20 +446,93 @@ export async function deactivateSessionBlock(
...
+  unlinkMulti.unlink(`s:${sessionId}:b:${sessionBlockId}:lb`)
+  activeInstanceIds.forEach((instanceId) => {
+    unlinkMulti.unlink(`s:${sessionId}:i:${instanceId}:info`)
+    unlinkMulti.unlink(`s:${sessionId}:i:${instanceId}:responseHashes`)
+    unlinkMulti.unlink(`s:${sessionId}:i:${instanceId}:responses`)
+    unlinkMulti.unlink(`s:${sessionId}:i:${instanceId}:results`)
+  })
+  unlinkMulti.exec()
```
rschlaefli, line 533: not awaiting here so the session is returned immediately

packages/graphql/src/services/sessions.ts
```diff
@@ -413,3 +549,42 @@ export async function getSession({ id }: { id: string }, ctx: ContextWithUser) {
...
+
+interface GetLeaderboardArgs {
+  sessionId: string
+}
+
+export async function getLeaderboard(
+  { sessionId }: GetLeaderboardArgs,
+  ctx: ContextWithUser
+) {
+  const top5 = await ctx.prisma.session
```
rschlaefli, line 561: simply return the top 5 for now, but the participant itself will also need to be displayed (and what if the participant is in the top 5)

packages/prisma/src/prisma/schema.prisma
```diff
@@ -188,6 +189,38 @@ model Course {
...
+  sessionBlock   SessionBlock? @relation(fields: [sessionBlockId], references: [id])
+  sessionBlockId Int?
+
+  session   Session? @relation(fields: [sessionId], references: [id])
+  sessionId String?  @db.Uuid
+
+  course   Course? @relation(fields: [courseId], references: [id])
+  courseId String? @db.Uuid
+
+  @@unique([type, participantId, sessionBlockId])
```
rschlaefli, line 219: three unique indices so that each of the use cases works: leaderboard of a session block, of a session, and of a course

packages/prisma/src/prisma/schema.prisma
```diff
@@ -231,9 +265,9 @@ enum AccessMode {
 model Session {
   id String @id @default(uuid()) @db.Uuid
 
-  isAudienceInteractionActive Boolean @default(false)
+  isAudienceInteractionActive Boolean @default(true)
```
rschlaefli, line 268: set these two to true by default (want to encourage their use)

PR #2878, opened 2022-09-14
feat(apps/frontend-pwa): add logic to login page and connect index to backend

florinavogel:


apps/frontend-pwa/src/components/CourseElement.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const computedClassNameButton = twMerge(
+    'h-full p-3 text-2xl rounded-r-md',
+    props.disabled ? 'bg-slate-400' : 'bg-slate-600'
+  )
+
+  return (
+    <div className="flex justify-between w-full mt-4 mb-4 mr-4align-center">
+      <span className="w-full p-4 text-l text-start rounded-l-md bg-uzh-grey-20 hover:bg-uzh-grey-40">
+        {/* TODO: ev. turn into link */}
+        {props.courseName}
```
rschlaefli: link to /course/\<ID> -> course overview with leaderboard, etc.

apps/frontend-pwa/src/components/CourseElement.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </span>
+      <span className={computedClassNameButton}>
+        <TogglePrimitive.Root
+          defaultPressed={subscribed}
+          onPressedChange={setSubscribed}
+          asChild
+        >
+          <button
```
rschlaefli, line 52: Design system button

apps/frontend-pwa/src/components/CourseElement.tsx
```diff
@@ -0,0 +1,79 @@
...
+            disabled={props.disabled}
+            onClick={() =>
+              props.onSubscribeClick(subscribed, setSubscribed, props.courseId)
+            }
+          >
+            {subscribed ? (
+              <FontAwesomeIcon
```
rschlaefli, line 59: \<Button.Icon> aus design system

apps/frontend-pwa/src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,19 @@
+import { faCircleXmark } from '@fortawesome/free-regular-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+type errorNotificationProps = {
+  message: string
+}
+
+function ErrorNotification(props: errorNotificationProps) {
```
rschlaefli, line 8: add to design system as Message component (this will be helpful everywhere)

apps/frontend-pwa/src/components/CourseElement.tsx
florinavogel: I'm not sure if this is a good idea since these are toggle buttons and are styled completely different than the design system button. When I just replace the button and leave the rest as is, this is what it would look like 
image

while it should look like this: 
image

PR #3150, opened 2022-12-10
feat: foundations of customizable achievements and rewards

rschlaefli:


packages/prisma/src/prisma/schema.prisma
```diff
@@ -245,8 +247,11 @@ model Course {
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
+  Title     Title[]
```
sjschlapbach: Does this property start with a capital letter on purpose?

PR #3187, opened 2022-12-30
enhance(cypress): rework selectors for data-cy, add draft workflow for GH actions

jnsgbl:
Tests are nearly updated to the new data-cy implementations.


jnsgbl:
All tests work now on the new data-cy implementations.

PR #3192, opened 2023-01-01
enhance(frontend-manage): display course join link and PIN through QR code

sjschlapbach:
This pull request introduces the following features:
- show the PIN required to join the course on the course overview page
- allow the user to directly offer a link to the course join page with pre-filled pin using a QR code

Screenshot 2023-01-01 at 21 51 37
Screenshot 2023-01-01 at 21 52 20
Screenshot 2023-01-01 at 21 54 45


apps/frontend-manage/src/pages/courses/[id].tsx
```diff
@@ -49,8 +52,39 @@ function CourseOverviewPage() {
...
+          </H1>
+          <div className="flex flex-row items-center gap-4 mb-2">
+            <QRPopup
+              relativeLink={`/course/${course.id}/join?pin=${course.pinCode}`}
+              triggerText="Kurs beitreten"
+              className={{ modal: 'w-[40rem]' }}
+            >
+              <H2>Kurs beitreten</H2>
+              <Link
+                href={`https://pwa.klicker.uzh.ch/course/${course.id}/join?pin=${course.pinCode}`}
```
Comment (rschlaefli) [inline apps/frontend-manage/src/pages/courses/[id].tsx]: TODO: dynamic base URL based on new env variable in .env.production (e.g., NEXT\_PUBLIC\_PWA\_URL=https://pwa.klicker.uzh.ch) and same for other environments

rschlaefli (approved):

apps/frontend-manage/src/pages/courses/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-manage/src/pages/courses/[id].tsx]: I added this directly to this PR so it that it can be used from here onwards and is not hard-coded over and over again and left for the end before the release.

PR #3249, opened 2023-01-17
style(frontend-manage): fix minor style issues on manage frontend

sjschlapbach:
This pull request addresses the following minor style issues / enhancements on the manage frontend:
- tags are no longer highlighted in red and the reset button indicates when it can be used and when not
- maximum one line of content but with markdown rendering is shown on the question pool question list
- it is now possible to also filter for sample solution and answer feedbacks using the tag filter menu


apps/frontend-manage/src/components/questions/TagList.tsx
```diff
@@ -42,8 +52,12 @@ function TagList({
       <div className="p-4 md:min-w-[17rem] border border-uzh-grey-60 border-solid md:max-h-full rounded-md h-max text-[0.9rem] overflow-y-auto">
         <Button
           className={{
-            root: 'w-full text-base bg-white hover:bg-grey-40 !py-[0.2rem] mb-1.5 flex flex-row items-center justify-center',
+            root: twMerge(
+              'w-full text-base bg-white hover:bg-grey-40 !py-[0.2rem] mb-1.5 flex flex-row items-center justify-center',
+              (activeTags.length > 0 || activeType) && theme.primaryText
```
sjschlapbach: add remaining boolean props as well that can be used to disable the button

PR #3255, opened 2023-01-21
enhance(frontend-manage): add basic tag management

sjschlapbach:
This pull request adds basic tag management and some bug fixes in connection with this:
- Tags can now be edited through the list on the left-hand side of the question pool
- Tags can now be deleted through the list on the left-hand side of the question pool
- Editing and Deleting tags is only possible when they are not actively selected in order to avoid race conditions
- Bugfix: When tags are created / modified on questions, the question pool view updates as to be expected

Screenshot 2023-01-21 at 13 26 42


apps/frontend-manage/src/components/questions/tags/TagActions.tsx
```diff
@@ -0,0 +1,58 @@
...
+  isDeletionModalOpen,
+  setIsDeletionModalOpen,
+}: TagActionsProps) {
+  const theme = useContext(ThemeContext)
+
+  return (
+    <div className="flex-row hidden text-black group-hover:flex">
+      <Button
+        basic
+        disabled={active}
```
sjschlapbach, line 30: Editing and deleting Tags is only possible when they are not active. Otherwise, strange race conditions are created and questions are filtered for tags that do not exist anymore in this form (and can therefore not even be de-selected).

PR #3256, opened 2023-01-22
enhance(frontend-manage): introduce session cancellation functionality

sjschlapbach:
This pull request re-adds the possibility to cancel sessions to Klicker v3.0


packages/graphql/src/services/sessions.ts
```diff
@@ -1583,3 +1583,99 @@ export async function getSessionEvaluation(
...
+  await Promise.all(
+    instances.map(async (instance) => {
+      await ctx.prisma.questionInstance.update({
+        where: {
+          id: instance.id,
+        },
+        data: {
+          participants: 0,
+          // TODO: figure out how to reset the results properly to initialValue stuff
+          results: {},
```
sjschlapbach, line 1668: The results object still has to be initialized correctly (probably using the `prepareInitialInstanceResults`-Funktion) for consistency reasons. Once this is fixed, cancelling sessions works again in principle.

packages/graphql/src/services/sessions.ts
```diff
@@ -1569,17 +1569,100 @@ export async function getSessionEvaluation(
...
+  await Promise.all(
+    instances.map(async (instance) => {
+      await ctx.prisma.questionInstance.update({
+        where: {
+          id: instance.id,
+        },
+        data: {
+          participants: 0,
+          // TODO: ensure that this typescript error does not point to a real potential issue
+          results: prepareInitialInstanceResults(instance.questionData),
```
sjschlapbach: This line create TypeScript errors at a potentially critical point. If this object is corrupted and code in other components relies on the default structure, this could lead to significant problems...

sjschlapbach: One TypeScript error originating from the initialization of the results object will be fixed later

PR #3259, opened 2023-01-23
chore(packages/prisma): merge seed data from submodule

BShaq:


packages/prisma/src/data/data/TEST.ts
```diff
@@ -0,0 +1,297 @@
...
+      {
+        correct: true,
+        feedback: 'Diese Aussage ist korrekt.',
+        value:
+          'Zusätzliche Sensitivitätsfaktoren erklären die Schwankung der Aktienrenditen besser.',
+      },
+    ],
+  },
+  {
+    id: 202,
```
rschlaefli:
```suggestion
    id: 4,
```

PR #3173, opened 2022-12-19
feat(apps/office-addin): add initial POC for content/taskpane add-in in Microsoft PowerPoint

florinavogel:
Screenshot from 2023-01-28 17-09-13

PR #3262, opened 2023-01-23
feat: introduce token-based login mechanism for controller app

sjschlapbach:
This pull request introduces a modified login mechanism for the controller app (frontend-control). This approach requires the user to enter his/her email address alongside a token, which can be generated in the manage frontend (limited validity of 10 minutes).
This login type is required for PR #3173 to enable login inside PPT without redirecting to external website (even when the user is usually logged in using SwitchAAI / Edu-ID

Screenshot 2023-01-23 at 17 19 29


packages/prisma/src/prisma/schema.prisma
```diff
@@ -53,6 +53,8 @@ model User {
   lastLoginAt         DateTime?
   deletionToken       String?
   deletionRequestedAt DateTime?
+  loginToken          String?
```
sjschlapbach, line 56: @rschlaefli These changes still require a prisma mutation for the database to be updated correctly. Other than that, this PR is ready to be merged as well.

sjschlapbach: Migration has been applied

PR #3284, opened 2023-01-27
feat(frontend-manage): enable searching and sorting in question pool

BShaq:


apps/frontend-manage/src/pages/index.tsx
```diff
@@ -124,22 +151,66 @@ function Index() {
...
-                          theme.primaryBgDark
-                        ),
-                      }}
-                      data={{ cy: 'create-question' }}
-                    >
-                      FRAGE ERSTELLEN
-                    </Button>
+                    <div className="flex flex-row content-center justify-between w-full h-full mt-4 pl-7">
+                      <div className="relative flex flex-row w-2.5/5 pb-6">
+                        <TextField
```
sjschlapbach, line 156: Even though it looks fine in the design-system, the icon passed to the TextField component seems to break something here...

Screenshot 2023-01-27 at 17 10 52

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -64,6 +71,26 @@ function Index() {
...
     useState(false)
 
+  const dropdownItems = [
+    { value: 'CREATED', label: 'Datum' },
+    { value: 'TITLE', label: 'Titel' },
+  ]
+
+  const [sortBy, setSortBy] = useState('')
+
+  const sortIcon = useMemo(() => {
```
sjschlapbach, line 81: Do we need 3 different icons? The list will initially be ordered in one way or the other anyway, so we could directly display this icon as an initial value. And then we can solve it inline at the TextField component with something like `sortBy?.asc ? faSortAsc : faSort` or similar?

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -64,6 +71,26 @@ function Index() {
...
+    if (!sortBy) {
+      return faSort
+    }
+
+    if (sort.asc) {
+      return faSortAsc
+    }
+
+    return faSortDesc
+  }, [sortBy, sort.asc])
```
sjschlapbach, line 91: To emphasize that we are sorting a list here, we could maybe use more explicit icons like faArrowUpWideShort or similar?

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -124,22 +151,66 @@ function Index() {
...
+                            handleSearch(newValue)
+                          }}
+                          icon={faMagnifyingGlass}
+                          className={{
+                            input: 'h-10',
+                            field: 'w-1/2',
+                          }}
+                        />
+                        <div className="flex flex-row justify-center w-1/2 ml-1">
+                          <Button
```
sjschlapbach: This button seems to shift backwards and forwards, depending on whether it is disabled or not?

Screenshot 2023-01-27 at 17 10 52
Screenshot 2023-01-27 at 17 10 37

sjschlapbach (changes requested):
Already works great :) Just found some minor cosmetic stuff that might be worth fixing.

apps/frontend-manage/src/pages/index.tsx
rschlaefli: image

On the table on the evaluation, we have the same behavior to show whether sorting is active on a column. While it is necessary there to distinguish which columns are sorted (and not here with basically "one column"), I would prefer using the same behavior and the same icons (see https://github.com/uzh-bf/design-system/blob/25e2d37166273a7d89af9ffbb5b72da230fd6fba/src/Table.tsx#L147).

So I suggest we leave this as implemented.

sjschlapbach: makes sense :)

apps/frontend-manage/src/pages/index.tsx
sjschlapbach: see above

apps/frontend-manage/src/pages/index.tsx
sjschlapbach: is fixed

apps/frontend-manage/src/pages/index.tsx
BShaq: As discussed with @rschlaefli yesterday, I applied a temporary fix for now. We'll have to investigate further later on.  @sjschlapbach Could you check again if it works now on your machine as well:

image

sjschlapbach: @BShaq Thanks, LGTM :) Works in all tested browsers

sjschlapbach (approved):

PR #3288, opened 2023-01-29
enhance: set-up persisted queries and CSRF

rschlaefli:
Implement CSRF and Persisted Queries as security measures. Persisted queries ensure that only queries that we have explicitly defined in our `ops` directory can be run against our backend (as the server accepts only hashes that exist in the mapping). This can be disabled in dev for better DX.


apps/frontend-control/src/lib/apollo.ts
```diff
@@ -8,8 +8,11 @@ import {
   split,
 } from '@apollo/client'
```
rschlaefli, line 9: all three apollo.ts are the same file -> extract this to shared-components at some point

packages/graphql/src/public/client.json
```diff
@@ -0,0 +1,81 @@
+{
```
rschlaefli, line 1: with this mapping, the client knows which hash to send to the server for which of the graphql ops

packages/graphql/src/public/server.json
```diff
@@ -0,0 +1,81 @@
+{
```
rschlaefli, line 1: this map helps the server look-up the specific query contents for the hash it received from the client

packages/graphql/tsup.config.ts
```diff
@@ -4,4 +4,5 @@ export default defineConfig({
   entry: ['src/index.ts', 'src/ops.ts', 'src/util.ts'],
   clean: true,
   dts: true,
+  publicDir: 'src/public',
```
rschlaefli, line 7: very useful -> allows us to automatically copy files from the `src/public` directory into `dist` on each build or dev generation (replacement for manual copy.js (could replace in `prisma` package as well). the client and server JSON are generated into the public folder and copied from there to dist.

PR #3301, opened 2023-02-02
feat(frontend-manage): enable configuration of colors for courses

BShaq:
This pull request introduces a new color picker that allows to configure the color of courses according to an individual's  preferences.


apps/frontend-manage/src/components/common/ColorPicker.tsx
```diff
@@ -0,0 +1,70 @@
...
+  color: string
+  onSubmit: (newColor: string) => void
+  onAbort: () => void
+}
+
+function ColorPicker({ color, onAbort, onSubmit }: Props) {
+  const theme = useContext(ThemeContext)
+  const [newColor, setNewColor] = useState(color)
+
+  const presetColors = ['#262FAD', '#016272', '#5FB1F9', '#FE7408', '#D84B39']
```
rschlaefli: Extract to constants.ts and import here to make it more accessible.

apps/frontend-manage/src/pages/courses/[id].tsx
```diff
@@ -21,18 +26,25 @@ function CourseOverviewPage() {
...
 
+  const [changeCourseColor] = useMutation(ChangeCourseColorDocument)
+
   useEffect(() => {
     if (data && !data.course) {
       router.push('/404')
     }
   }, [data, router])
 
+  const toggleColorPicker = useCallback(() => {
```
Comment (rschlaefli) [inline apps/frontend-manage/src/pages/courses/[id].tsx]: Is there a specific reason for useCallback here? I think we could just call it like setIsColorPickerVisible(prevState => !prevState)

package.json
```diff
@@ -61,7 +61,8 @@
   },
   "dependencies": {
     "@prisma/client": "4.8.1",
-    "prisma": "4.8.1"
+    "prisma": "4.8.1",
+    "react-colorful": "^5.6.1"
```
rschlaefli: Always install with a pinned version tag (no `^`), except @types/X.

So `npm install -E react-colorful` or `npm install -E -D react-colorful`. This works better with our upgrade approach (explicitly upgrade versions, not `npm update`).

-> Remove `^` and run `npm install` to regenerate lockfile

rschlaefli: Also, `react-colorful` is a dependency of `apps/frontend-manage`, not the root project (this would break in builds or production)! Remove from the root `package.json` and either `cd apps/frontend-manage` and `npm install -E react-colorful`, or run `npm i -E -w frontend-manage react-colorful` (or something like this, not sure, but -w specified which workspace to install the package in).

packages/graphql/src/nexus.ts
```diff
@@ -1979,6 +1979,19 @@ export const Mutation = objectType({
         return ParticipantGroupService.submitGroupActivityDecisions(args, ctx)
       },
     })
+
+    t.field('changeCourseColor', {
+      type: Course,
+      args: {
+        courseId: nonNull(idArg()),
+        color: nonNull(
```
rschlaefli: Run `npm run format` and commit separately, I think prettier would put this on the same line

rschlaefli (changes requested):
Thanks, almost perfect ;) Requested some small changes, and please reformat (commit separately).

apps/frontend-manage/src/pages/courses/[id].tsx
Comment (BShaq) [reply apps/frontend-manage/src/pages/courses/[id].tsx]: I saw this in multiple similar examples in other projects. As far as I understood, this use of the hook could be seen as a performance optimization, since otherwise the `toggleColorPicker` function would be unnecessarily re-created on every re-render. This way, it will be cached and only updated if there is a change in its dependencies.  

But I also have no issue with leaving it out if you wish :)

apps/frontend-manage/src/components/common/ColorPicker.tsx
sjschlapbach: @BShaq There is a constants file in the `shared-components` package, which we can  use to have access to those variables in all different frontends :) You can find it at `packages/shared-components/src/constants.ts`

packages/shared-components/.prettierignore
```diff
@@ -0,0 +1 @@
+src/tailwind.css
```
sjschlapbach, line 1: @BShaq Just fyi - I updated the two of the prettier ignore files so that `npm run format` does not format automatically generated files. Otherwise, we will always have 500+ changes for example for the tailwind file in `shared-components` alone. And these will go back and forth with each new generate of the file...

sjschlapbach: This was btw. also the reason that the sonarcloud analysis returned 91 code smells instead of 7 😅

apps/frontend-manage/src/components/common/ColorPicker.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <div className="absolute flex flex-col p-1 bg-white rounded-md shadow-md w-80 left-10 top-8">
+      <HexColorPicker
+        style={{ width: 'auto' }}
+        color={newColor}
+        onChange={setNewColor}
+      />
+      <div className="grid grid-cols-5 gap-0.5 pt-3 pb-3 justify-items-center">
+        {PRESET_COURSE_COLORS.map((presetColor, index) => (
+          <button
```
sjschlapbach: Is there a specific reason not to use our custom <Button> component here (just for consistency)?

sjschlapbach (commented):
The color picker looks great 😄

apps/frontend-manage/src/components/common/ColorPicker.tsx
BShaq: @sjschlapbach Thanks for the hint! Adjusted the code accordingly :)

sjschlapbach: Perfect 👍

packages/shared-components/.prettierignore
BShaq: Oh, thanks for pointing that out :)

apps/frontend-manage/src/components/common/ColorPicker.tsx
BShaq: Adjusted it ✅

sjschlapbach: Thanks 👍

rschlaefli (approved):

PR #3307, opened 2023-02-03
chore: update design-system version with fixed select component

sjschlapbach:
This pull requests adds an updated design-system component, which fixed the auto-selection problem.


BShaq (approved):
LGTM :)

PR #3325, opened 2023-02-09
enhance(frontend-manage): separate toggle functionality for live Q&A and confusion feedback

BShaq:
This PR addresses feedback from users who wish to decide which method of audience interaction they want to use during a live-session, i.e., only live Q&A, only confusion feedback or both.


apps/frontend-manage/src/components/interaction/AudienceInteraction.tsx
```diff
@@ -71,173 +73,212 @@ function AudienceInteraction({
...
+        )}
+      >
+        <div className="flex flex-col w-2/3 pr-5">
+          <div className="flex flex-row justify-between">
+            <div className="flex pr-5 text-2xl font-bold">Live Q&A</div>
+            <div className="flex flex-row gap-6">
+              <Switch
+                checked={isLiveQAEnabled}
+                onCheckedChange={(): void => {
+                  changeSessionSettings({
```
sjschlapbach, line 102: Only because we turn of the liveQAChannel we shouldn't turn off moderation automatically with it, right? The button is already disabled, but why should we reset the button to our default? And the setting won't have an effect anyways as long as the QA Channel is disabled

apps/frontend-pwa/src/components/liveSession/FeedbackArea.tsx
```diff
@@ -80,7 +80,15 @@ function Subscriber({ subscribeToMore, sessionId }) {
   return <div></div>
 }
```
sjschlapbach, line 82: On the student page, there is still a bug in the state logic, which could possibly also be fixed with this PR.

If the window size is decreased to mobile, the user switches to the feedback view and then increases the window size again, the question does not appear in a side-by-side view as it should.

packages/graphql/src/services/sessions.ts
```diff
@@ -1123,6 +1127,8 @@ export async function changeSessionSettings(
     where: { id },
     data: {
       isAudienceInteractionActive: isAudienceInteractionActive ?? undefined,
```
sjschlapbach: As the isAudienceInteractionActive prop is not used anymore and will be dropped from the database, we should make sure that no DB service / frontend component still checks it 😅

sjschlapbach (changes requested):

packages/prisma/src/prisma/migrations/20220925181035_initial_migration/migration.sql
```diff
@@ -182,6 +182,8 @@ CREATE TABLE "SessionBlock" (
 CREATE TABLE "Session" (
     "id" UUID NOT NULL,
     "isAudienceInteractionActive" BOOLEAN NOT NULL DEFAULT false,
+    "isLiveQAEnabled" BOOLEAN NOT NULL DEFAULT false,
```
rschlaefli: Remove this, migrations are not written by hand (with Prisma CLI, will do this at the end before merge :))

apps/frontend-manage/src/components/interaction/AudienceInteraction.tsx
BShaq: done ✅

apps/frontend-pwa/src/components/liveSession/FeedbackArea.tsx
BShaq: fixed ✅

packages/graphql/src/services/sessions.ts
BShaq: done ✅

packages/prisma/src/prisma/migrations/20220925181035_initial_migration/migration.sql
BShaq: reverted✅

rschlaefli (approved):

apps/frontend-pwa/src/components/liveSession/FeedbackArea.tsx
sjschlapbach: Great, thanks! this has been a bug for a long time already! 😄

PR #3327, opened 2023-02-13
fix(frontend-manage): fix NaN values in tables and bar charts

BShaq:
This PR fixes the NaN values that appear for SC, MC and KPRIM questions on the table chart when no answer has yet been submitted.


sjschlapbach (commented):
LGTM :) @BShaq Does a similar issues also occur for NR and/or FT questions when there are no answers yet? Then it might be worth to also fixed this right away on the same branch. And do the other evaluation chart components (bar chart, etc. handle the case of missing data correctly - or the histogram for numerical questions)?

rschlaefli (approved):

BShaq:
> LGTM :) @BShaq Does a similar issues also occur for NR and/or FT questions when there are no answers yet? Then it might be worth to also fixed this right away on the same branch. And do the other evaluation chart components (bar chart, etc. handle the case of missing data correctly - or the histogram for numerical questions)?

Thanks for your feedback  @sjschlapbach! 

Regarding your first question - I did not come across such difficulties for NR and FT questions in relation with the table. AFAIK, there are  no predefined answers for such questions where the students have to pick one/multiple from. This, in turn, means that as long as there are no answers submitted the table will be empty - which I think is as intended, right?

As for the other charts, I only found one issue with the bar chart in context with NR questions. If there are no answers, the x-axis is then labeled with `0` and `auto`, as shown in the screenshot below:

image

In this case, I'd opt for a solution like in my latest [commit](https://github.com/uzh-bf/klicker-uzh/pull/3327/commits/97f40ab074ed639ff91a8bd474c8d20f80d301cc) where we explicitly set the x-axis to `0` if there are no answers yet: 

image

What do you think of this suggestion?

sjschlapbach (approved):

rschlaefli:
Simplified the BarChart a bit, merging this now :)

PR #3346, opened 2023-02-14
fix(apps/frontend-*): several fixes for kprim questions

BShaq:


apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -89,18 +90,28 @@ function LearningElement({ courseId, id }: Props) {
...
 
+  const formatResponse = () => {
+    if (
+      questionData?.type === QuestionType.SC ||
+      questionData?.type === QuestionType.MC
+    ) {
+      return { choices: response as number[] }
+    } else if (questionData?.type === QuestionType.KPRIM) {
+      let choicesArray: number[] = []
+      Object.keys(response).map((key) => {
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: refactor so it does not use let and array.push (mutable way)

-> naive approach: .filter and then .map, store the returned value of .map in choicesarray
-> better approach: use .flatMap to combine .filter and .map (return [] in flatMap to skip the element, [value] to add it to the output). flatMap combines all of the returned arrays into one large array, so empty arrays -> filtered out

Comment (rschlaefli) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: also, add questionData and response etc. as an argument to formatResponse and place it as a function outside of the react component (cleaner and better reusability). best approach would be a custom hook, but not necessary here

packages/shared-components/src/constants.ts
```diff
@@ -31,6 +31,11 @@ export const CHART_COLORS: string[] = [
   'rgb(236, 170, 56)', // 9
 ]
 
+export const CHART_SOLUTION_COLORS = {
```
rschlaefli, line 34: good idea!

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -53,10 +53,11 @@ function LearningElement({ courseId, id }: Props) {
     if (questionData?.type) {
       if (
         questionData.type === QuestionType.SC ||
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: import the QuestionType from @klicker-uzh/graphql/dist/ops to reuse the graphql type instead of our own const (we can gradually get rid of our own const by using the generated types from the backend). use it as QuestionType.Sc etc. (type generation changes .SC to .Sc, but the string stored within is all caps and still works)

rschlaefli (changes requested):
minimal changes to code style, otherwise LGTM

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (BShaq) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: thanks for the feedback! I adjusted the code accordingly 👍

rschlaefli (approved):

PR #3347, opened 2023-02-15
chore: add initial export and import scripts for SC/MC questions (v2->v3)

rschlaefli:


packages/prisma/src/data/importQuestions.ts
```diff
@@ -0,0 +1,112 @@
...
+                feedback: '<br>',
+              }
+            }
+          ),
+        }
+      } else if (question.type === 'NR') {
+        // TODO
+        throw new Error('Unsupported question type (NR)')
+      } else if (question.type === 'FT') {
+        // TODO
```
sjschlapbach, line 77: Is there a specific reason that we do not offer the migration for NR and FT questions yet?

sjschlapbach (approved):
LGTM :)

packages/prisma/src/data/importQuestions.ts
rschlaefli: Just that there was no user with NR and FT to migrate yet :)

PR #3352, opened 2023-02-17
feat: flag questions

rschlaefli:


sjschlapbach (commented):
Mutation looks perfect to me :)

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -22,11 +24,23 @@ import {
...
+} from '@uzh-bf/design-system'
 import dayjs from 'dayjs'
+import { Form, Formik } from 'formik'
 import { GetServerSideProps } from 'next'
 import { useRouter } from 'next/router'
-import { useEffect, useMemo, useState } from 'react'
+import { useContext, useEffect, useMemo, useState } from 'react'
+import { twMerge } from 'tailwind-merge'
+import * as Yup from 'yup'
+import ContentInput from '../../../../../../frontend-manage/src/components/questions/ContentInput'
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: Import from frontend-manage will fail during build / in production

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -42,14 +56,18 @@ const ORDER_TYPE_LABEL = {
 interface Props {
   courseId: string
   id: string
+  description: string
+  submitText: string
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: Neither description nor submitText are provided to the page through the props function at the end of the file. Therefore, these values either have to be removed or fetched from somewhere else.

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -323,7 +371,86 @@ function LearningElement({ courseId, id }: Props) {
                           <div className="font-bold">Multiplikator</div>
                           <div>{currentInstance.pointsMultiplier}x</div>
                         </div>
-                        <FontAwesomeIcon icon={faFlag}></FontAwesomeIcon>
+                        <Modal
+                          className={{ content: 'm-10 md:m-20' }}
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: We should style the modal a bit more to only take the space it needs and still work nicely on mobile devices. Probably this will also need testing in production later on.

Screenshot 2023-02-17 at 16 44 12

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -42,14 +56,18 @@ const ORDER_TYPE_LABEL = {
...
 
 // TODO: leaderboard and points screen after all questions have been completed?
 // TODO: different question types (FREE and RANGE)
-function LearningElement({ courseId, id }: Props) {
+function LearningElement({ courseId, id, description, submitText }: Props) {
   const router = useRouter()
   const [response, setResponse] = useState<{} | number[] | string | null>(null)
   const [currentIx, setCurrentIx] = useState(-1)
+  const [isOpen, setIsOpen] = useState(false)
+  const [open, setOpen] = useState(false)
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: We should probably rename those states so that it is clear from the respective name, which one is used for the Modal and which one is used for the Toast.

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
```diff
@@ -62,6 +80,36 @@ function LearningElement({ courseId, id }: Props) {
     skip: !router.query.courseId,
   })
 
+  const theme = useContext(Theme)
+  const [flagQuestion] = useMutation(FlagQuestionDocument, {})
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: We probably don't need to pass an empty object here?

sjschlapbach (changes requested):
Already looks very nice and works perfectly, only some smaller things that need to be fixed in order for the build to succeed and to provide optimal readability for the code :)

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: removed unused arguments

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: renamed states

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: removed

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: resolved

apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-pwa/src/pages/course/[courseId]/element/[id].tsx]: component has been moved to shared-components, but replaced by textarea in the end

sjschlapbach (approved):

PR #3358, opened 2023-02-17
feat: add question level explanation

sjschlapbach:
This pull request adds the possibility to add a explanation on question level, which is displayed in learning elements and micro session questions. It can be used to define a general explanation, how the correct solution can be obtained, without providing feedback for every single answering option.


sjschlapbach (commented):
@rschlaefli We need to apply a db migration here as well before merging this branch to add the required explanation attribute on the question

rschlaefli (approved):

PR #3361, opened 2023-02-18
feat: add support for NR questions in learning elements

sjschlapbach:
Screenshot 2023-02-19 at 18 53 38


apps/frontend-pwa/src/components/common/OptionsDisplay.tsx
```diff
@@ -276,7 +276,12 @@ export function Options({
             max={options?.restrictions?.max}
             value={response}
             onChange={onChangeResponse}
-            valid={isResponseValid}
+            valid={
+              isResponseValid &&
+              ((parseFloat(response) >= options?.restrictions?.min &&
+                parseFloat(response) <= options?.restrictions?.max) ||
+                response === '')
```
rschlaefli: How is an empty string valid for a number input at any time?

apps/frontend-pwa/src/components/common/OptionsDisplay.tsx
```diff
@@ -340,7 +345,11 @@ function OptionsDisplay({
               response?.length === 0) ||
```
rschlaefli, line 377: Should probably move this lengthy logic to a function

apps/frontend-pwa/src/components/evaluation/EvaluationDisplay.tsx
```diff
@@ -92,7 +102,30 @@ function EvaluationDisplay({ options, questionType, evaluation }: Props) {
       return <div></div>
 
     case QuestionType.NUMERICAL:
-      return <div></div>
+      const results = Object.entries(
+        evaluation.answers as Record<string, number>
```
rschlaefli, line 106: I thought we had no answers object?

packages/shared-components/package.json
```diff
@@ -30,7 +30,9 @@
     "tailwind-merge": "^1.8.0",
     "tailwindcss": "^3.2.4",
     "typescript": "^4.9.3",
-    "yup": "^1.0.0"
+    "yup": "^1.0.0",
+    "recharts": "^2.4.1",
+    "lodash": "^4.17.21"
```
rschlaefli, line 35: Why lodash? Try to use ramda, so we can prevent adding another dependency

packages/shared-components/src/Histogram.tsx
```diff
@@ -242,27 +264,32 @@ function Histogram({
...
-      <div className="flex flex-row items-center float-right gap-2 mr-4">
-        <div className="font-bold">Bins:</div>
-        <input
-          className="rounded-md"
-          type="number"
-          value={numBins}
-          onChange={(e) => setNumBins(+e.target.value)}
-        />
-      </div>
+      {!hideBins && (
```
rschlaefli, line 282: This is very specific to the evaluation. Long-term, we should probably have the Histogram only get numBins from props, and extract and manage this on the evaluation components.

rschlaefli (approved):

PR #3362, opened 2023-02-18
feat: image embedding, grid layout, editing preview, numerical questions in learning elements, and many more

rschlaefli:
- support grid layout in learning elements and microlearning
- implement better approach for image embedding and enlargement in modal (apply everywhere)
- fix /join/<shortname> page
- add tab for course information on /course/<id> that is only shown if course description is set
- only show group creation/join tab if not already participating in a group
- hide generic question explanation if it is not specified
- support image embedding in table on evaluation
- fix height of chart overlapping with footer (z-index on footer and padding instead of margin for bottom)
- use next/dynamic to prevent SSR issues where Markdown component is used (and getServerSideProps)
- consolidate question edit and preview modal
- simplify the layout on question pool (remove id, rearrange question contents)
- refactor evaluation layout to flex-based and fully flexible (with fixed question collapsible)
- rework tailwind to separate modules for utilities and components
- extract Ellipsis into markdown component
- update layout on login
- allow additional parameters for numerical questions (accuracy, unit) and enforce on student view
- basic link support within markdown
- validate question types for support on creation wizard (e.g., no KPRIM in live sessions, no FREE_TEXT on learning elements or micro sessions)


apps/frontend-manage/src/components/questions/Question.tsx
```diff
@@ -92,134 +91,120 @@ function Question({
...
-              className={{ root: 'justify-center h-10 bg-white w-36' }}
-              onClick={(): void => setIsDuplicationModalOpen(true)}
-              // TODO: implement
-              disabled
-            >
-              Duplizieren
-            </Button>
-            {/* {isDuplicationModalOpen && (
+        <div className="flex flex-col md:flex-row md:justify-between">
+          <div></div>
```
sjschlapbach: do we need this empty div here?

apps/frontend-manage/src/components/questions/QuestionPreviewModal.tsx
```diff
@@ -1,107 +0,0 @@
-import { useQuery } from '@apollo/client'
```
sjschlapbach, line 1: Are we sure that we want to delete the preview modal completely? This would allow the lecturers to test the exact behavior of the student component incl. disabled buttons etc. with specified restrictions

apps/frontend-pwa/src/pages/course/[courseId]/index.tsx
```diff
@@ -88,20 +94,23 @@ function CourseOverview({ courseId }: any) {
...
+            )}
           </Tabs.TabList>
 
-          <Tabs.TabContent key="course" value="global" className="md:px-4">
-            {course.description && (
-              <div className="p-4 mb-4 border rounded bg-slate-100">
-                <Markdown content={course.description} />
-              </div>
-            )}
+          {course?.description && (
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/index.tsx:114]: For some reason, this tab is also shown, when the description is "null" in the database:
Additionally, we would have to consider the case where the user started editing the field as it might contain only break and newline characters, where this test would fail as well...
(image for description null)
Screenshot 2023-02-19 at 12 31 19

apps/frontend-manage/src/pages/sessions/[id]/evaluation.tsx
```diff
@@ -207,9 +207,9 @@ function Evaluation() {
 
       <div
         className={twMerge(
-          'flex justify-between flex-col mt-11 [height:_calc(100%-6.25rem)] mb-14',
+          'flex justify-between flex-col mt-11 [height:_calc(100%-6.25rem)] pb-14',
           (showFeedbacks || showConfusion || showLeaderboard) &&
-            '[height:_calc(100%-7.25rem)] mb-18'
+            '[height:_calc(100%-7.25rem)] pb-18'
```
Comment (sjschlapbach) [inline apps/frontend-manage/src/pages/sessions/[id]/evaluation.tsx]: Removing the margin and adding a padding seems to lead to issues in some cases with a white bar inside the evaluation area. We should investigate, if the margin did indeed work better or if there is some other kind of issue as well.

Screenshot 2023-02-19 at 12 39 08
Screenshot 2023-02-19 at 12 45 28

Comment (sjschlapbach) [reply apps/frontend-manage/src/pages/sessions/[id]/evaluation.tsx]: Issue does not occur, when the questions are larger and displayed with a collapsible

apps/frontend-manage/src/components/questions/Question.tsx
sjschlapbach: fixed

sjschlapbach (approved):

PR #3403, opened 2023-02-24
enhance: award xp during live sessions

sjschlapbach:
This pull request will add support for xp awarding during live sessions.


sjschlapbach (commented):
@rschlaefli This bit of code already seems to do the job? Could you maybe double-check it when you have time, and if it makes sense and if so, mark it as ready to review / reviewed?

rschlaefli (approved):

PR #3432, opened 2023-03-04
fix: update validation logic for numerical questions

sjschlapbach:


sjschlapbach (commented):
@rschlaefli If we want to ensure also on the backend side that only responses in the specified range can be submitted, the `backend-response-processor` probably also needs a few more lines of code. Until now, this is only ensured for learning elements and micro sessions through the corresponding mutation that is called there.

packages/graphql/src/services/learningElements.ts
```diff
@@ -219,9 +219,14 @@ export async function respondToQuestionInstance(
       }
 
       case QuestionType.NUMERICAL: {
-        if (!response.value) {
-          break
+        if (
+          !response.value ||
```
rschlaefli:
```suggestion
          typeof response.value === 'undefined' ||
```

packages/shared-components/src/StudentQuestion.tsx
```diff
@@ -129,15 +129,20 @@ export const StudentQuestion = ({
     }
   }
 
-  const onNumericalValueChange = (inputValue: any): void => {
-    const inputEmpty =
-      inputValue !== 0 && (!inputValue || inputValue.length === 0)
+  const onNumericalValueChange = (inputValue: string): void => {
+    const inputEmpty = !inputValue || inputValue.length === 0
```
rschlaefli:
```suggestion
    const inputEmpty = typeof inputValue === 'undefined' || inputValue === ''
```

rschlaefli (changes requested):

packages/graphql/src/services/learningElements.ts
sjschlapbach: Shouldn't be an issue for strings, but I still adjusted it to test for undefined values, empty strings and null values as well (just to be sure 😄)

packages/shared-components/src/StudentQuestion.tsx
sjschlapbach: Same as above

rschlaefli (approved):

PR #3445, opened 2023-03-09
enhance: implement derived levels and display collected XP after learning element completion

sjschlapbach:


packages/graphql/src/schema/participant.ts
```diff
@@ -92,8 +94,23 @@ export const Participant = ParticipantRef.implement({
     }),
 
     xp: t.exposeInt('xp'),
-    level: t.exposeInt('level'),
-    levelData: t.expose('levelData', { type: LevelRef, nullable: true }),
+    level: t.int({
```
sjschlapbach, line 97: @rschlaefli This PR will need a migration before merging to drop the "level" and "levelData" columns on the participant table. Both of them are now directly derived from the number of xp through the participant resolver.

rschlaefli (approved):

PR #3462, opened 2023-03-14
enhance(apps/frontend-manage): implement folder question pool view

sjschlapbach:


sjschlapbach:
As extensions of this are not future-proof based on the current approach, this PR will be closed

PR #3489, opened 2023-03-25
fix(apps/frontend-manage): ensure that sorting of tables works as to be expected

sjschlapbach:
This pull request uses an updated version of the design-system table component, which supports both transformers and formatters. This allows to use the markdown component to display elements correctly, while still using the sorting functionality, which was broken for these cases before.


apps/frontend-manage/src/components/evaluation/TableChart.tsx
```diff
@@ -25,27 +25,23 @@ function TableChart({
 }: TableChartProps): React.ReactElement {
   const ref = useRef<{ reset: () => void }>(null)
 
+  const formatCorrect = (value: boolean | undefined | null): string => {
```
rschlaefli: maybe we can extract formatters to shared components?

apps/frontend-manage/src/components/evaluation/TableChart.tsx
```diff
@@ -25,27 +25,23 @@ function TableChart({
...
+    if (value === true) return 'T'
```
rschlaefli: should we translate this? or use icons instead of T/F?

apps/frontend-manage/src/components/evaluation/TableChart.tsx
```diff
@@ -25,27 +25,23 @@ function TableChart({
...
             count: data.results[index].count,
-            value: (
-              <Markdown
-                content={choice.value}
-                className={{ img: 'max-h-32' }}
-              />
-            ),
-            correct: choice.correct ? 'T' : 'F',
+            value: choice.value,
+            correct: formatCorrect(choice.correct),
```
rschlaefli: why is the formatting done on this level and not passed through as a formatter for the correct column? now that we have this abstraction? :)

apps/frontend-manage/src/components/evaluation/TableChart.tsx
```diff
@@ -25,27 +25,23 @@ function TableChart({
...
             percentage: data.participants
-              ? String(
-                  (
-                    (data.results[index].count / data.participants) *
-                    100
-                  ).toFixed()
-                ) + ' %'
-              : '0 %',
+              ? data.results[index].count / data.participants
```
rschlaefli, line 44: this would be a case for a transformer?

rschlaefli: transformers and formatters should get the value of the column first, but also access to the entire row, so that such calculations can be done within them and not on a higher-level.

rschlaefli (changes requested):

apps/frontend-manage/src/components/evaluation/TableChart.tsx
sjschlapbach: I'm currently not aware of any other function where we need this formatter? But yes, as soon as we need it somewhere else as well, this would definitely be a good idea.

apps/frontend-manage/src/components/evaluation/TableChart.tsx
sjschlapbach: icons would definitely be a nice idea, will be implemented

apps/frontend-manage/src/components/evaluation/TableChart.tsx
sjschlapbach: this wasn't possible until now, because of type constraints that were only lifted with the most recent design-system release. Will be changed in this PR

apps/frontend-manage/src/components/evaluation/TableChart.tsx
sjschlapbach: done

apps/frontend-manage/src/components/evaluation/TableChart.tsx
sjschlapbach: done

rschlaefli (approved):

PR #3555, opened 2023-04-18
fix: ensure that only integer point amounts are collectable

sjschlapbach:


sjschlapbach:
@rschlaefli This pull request still needs a db migration for the score type change from `Float` to `Int` to be effective.

rschlaefli (approved):

PR #3574, opened 2023-04-27
enhance: automate push notifications and implement unsubscribe functionality

BShaq:
This PR introduces several improvements to our current push notification implementation with a focus on automation, security and user preferences.

- The main feature introduced is the automation of push notifications. This change moves away from manually executing scripts and towards using a newly created API endpoint that handles the sending of push notifications. This is more efficient and consistent.
- Complementing the new API endpoint, a new cron job manifest has been added that sets up a regular trigger for the endpoint, which ensures that push notifications are sent every 5 minutes. Therefore, this guarantees that our users are informed as soon as possible whenever a new microsession is published.
- To ensure the security of the endpoints that are triggered by cron jobs, a token-based authentication for the `sendPushNotifications` and `updateAverageGroupScores` endpoints has been introduced. This measure prevents unauthorized access to these endpoints.
- Finally, this PR also enhances the user experience by giving participants the ability to unsubscribe from push notifications.


apps/frontend-pwa/src/pages/index.tsx
```diff
@@ -52,6 +54,8 @@ const Index = function () {
   }, [])
 
   const [subscribeToPush] = useMutation(SubscribeToPushDocument)
+  const [unsubscribeFromPush] = useMutation(UnsubscribeFromPushDocument)
+  const [sendPushNotifications] = useMutation(SendPushNotificationsDocument)
```
rschlaefli: this should be removed before the merge, as we do not want to allow triggering push from frontend

packages/graphql/src/graphql/ops/MSendPushNotifications.graphql
```diff
@@ -0,0 +1,3 @@
+mutation SendPushNotifications {
+  sendPushNotifications
```
rschlaefli, line 2: this will need to accept a secret as a parameter to ensure it can only be triggered from our own cronjon

rschlaefli, line 2: this will need to accept a secret as a parameter to ensure it can only be triggered from our own cronjon

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +51,130 @@ export async function subscribeToPush(
...
+    })
+    return true
+  } catch(error) {
+    console.error("An error occured while trying to unsubscribe from push notifications: ", error)
+    return false
+  }
+}
+
+export async function sendPushNotifications( ctx: Context) {
+  webpush.setVapidDetails(
```
rschlaefli, line 95: before doing anything, compare the secret passed as an argument and compare it to a secret token from process.env (e.g., the PUSH\_SECRET)

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +51,130 @@ export async function subscribeToPush(
...
+          subscriptions: true,
+        },
+      },
+      instances: false,
+    },
+  })
+
+
+  await Promise.all(
+    microSessions.map(async (microSession: MicroSession) => {
```
rschlaefli, line 126: we will need to think about scaling of this. could it lead to issues down the line when many people create a lot of micro sessions? we should balance the execution count of the cronjob (e.g., every 5 minutes) against the number of created micro sessions per hour or similar

rschlaefli: it could potentially be extracted into a background process

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +51,130 @@ export async function subscribeToPush(
...
+      } catch (error) {
+        console.error("An error occured while trying to send the push notifications: ", error)
+      }
+    })
+ )
+
+  return true
+}
+
+function sleep(ms) {
```
rschlaefli: generic function => extract to utils and add types

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +51,130 @@ export async function subscribeToPush(
...
+        {
+          endpoint: sub.endpoint,
+          keys: {
+            auth: sub.auth,
+            p256dh: sub.p256dh,
+          },
+        },
+        JSON.stringify({
+          message:
+            'Das Microlearning für BFI Woche 5 ist bis morgen um 09:00 verfügbar.',
```
rschlaefli: the message needs to be parametrized to reflect the name of the course and the name of the microlearning, as well as the timing. anything else that the lecturer should be able to customize?

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +51,130 @@ export async function subscribeToPush(
...
+  return true
+}
+
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+async function sendPushNotificationsToSubscribers(subscriptions: PushSubscription[], ctx: Context) {
+  for (let sub of subscriptions) {
+    await sleep(500)
```
rschlaefli: sleep should not be used, instead rely on a clean approach using only promises

packages/graphql/src/services/notifications.ts
rschlaefli: also, when adding a variable to process.env, also remember to add it to doppler, as well as to the deployment code (in deploy/) -> ask RS for specifics on this

packages/graphql/src/services/notifications.ts
BShaq: added a corresponding TODO

packages/graphql/src/graphql/ops/MSendPushNotifications.graphql
BShaq: For documentation: we decided to pass the token as an HTTP header.

BShaq: For documentation: we decided to pass the token as an HTTP header.

apps/frontend-pwa/src/pages/index.tsx
```diff
@@ -147,62 +149,119 @@ const Index = function () {
     return <div>loading...</div>
   }
 
-  async function onSubscribeClick(subscribed: boolean, courseId: string) {
+  async function onSubscribeClick(
```
rschlaefli, line 179: we should think about moving these generic hooks to shared components, as we might implement push on other PWA as well. also, would make this component much cleaner. the mutation to actually execute could be passed in as a param to the hook to make it generic.

packages/graphql/src/lib/util.ts
```diff
@@ -8,4 +11,35 @@ export function shuffle<T>(array: Array<T>): Array<T> {
...
+  if (typeof process.env.CRON_TOKEN === 'undefined') {
+    throw new GraphQLError("No token available.")
+  }
+  
+  if (ctx.req.headers["x-token"] !== process.env.CRON_TOKEN) {
+    throw new GraphQLError("Token not valid.")
+  }
+}
+
+export function formatDate(dateTime: Date) {
```
rschlaefli, line 32: should not work with js dates (too much of a pain). instead, use dayjs to format the date which is already installed

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +52,121 @@ export async function subscribeToPush(
...
+    })
+    return true
+  } catch(error) {
+    console.error("An error occured while trying to unsubscribe from push notifications: ", error)
+    return false
+  }
+}
+
+export async function sendPushNotifications( ctx: Context) {
+  webpush.setVapidDetails(
```
rschlaefli, line 95: check beforehand if VAPID\_PUBLIC\_KEY and VAPID\_PRIVATE\_KEY are set. if not, return early. also, the support email used here should be customizable using an env variable (like we did with the cron token).

rschlaefli, line 95: check beforehand if VAPID\_PUBLIC\_KEY and VAPID\_PRIVATE\_KEY are set. if not, return early. also, the support email used here should be customizable using an env variable (like we did with the cron token).

packages/graphql/src/services/notifications.ts
```diff
@@ -49,3 +52,121 @@ export async function subscribeToPush(
...
+        {
+          endpoint: sub.endpoint,
+          keys: {
+            auth: sub.auth,
+            p256dh: sub.p256dh,
+          },
+        },
+        JSON.stringify({
+          message:
+            `Das Microlearning ${microSession.displayName} für ${microSession.course.displayName} ist bis am ${formattedDate.date} um ${formattedDate.time} Uhr verfügbar.`,
```
rschlaefli: we will need to think about the translation of this when we use multilanguage. do we need to store the language of a course on the course entity and use it here to format the string? what about multilanguage courses? do we store the language on the user subscription (e.g., send it to the backend from the current frontend language when subscribing) and use this language when sending to the specific user?

rschlaefli (changes requested):

PR #3614, opened 2023-05-12
enhance(frontend-manage): convert frontend-manage to PWA

BShaq:


apps/frontend-manage/src/pages/_document.tsx
```diff
@@ -0,0 +1,253 @@
+// @ts-nocheck
+
+import Document, { Head, Html, Main, NextScript } from 'next/document'
+
+class MyDocument extends Document {
```
rschlaefli: refactor to functional component instead of class (even if we lose some typescript stuff)

https://nextjs.org/docs/pages/building-your-application/routing/custom-document

apps/frontend-manage/src/pages/_offline.tsx
```diff
@@ -0,0 +1,15 @@
+import { UserNotification } from '@uzh-bf/design-system'
+import Layout from '../components/Layout'
+
+const fallback = () => (
```
rschlaefli: can we move this to shared components and use it across all PWA?

rschlaefli (changes requested):

sjschlapbach:
@BShaq PR #3625 implements the standardized layout for all login pages, including installation hints for PWA applications. As this will probably also become relevant for this PR, you should be able to simply set `installationHint` in the `LoginForm` component to `true` in order to display the installation hint for PWAs after the conversion is completed :)

rschlaefli (approved):

PR #3624, opened 2023-05-21
chore: upgrade all packages to the latest version

rschlaefli:


sjschlapbach:
@rschlaefli On MacOS the upgrades currently cause the database resetting to fail. `npm run prisma:reset` returns the following error:
```
Running generate... (Use --skip-generate to skip the generators)
assertion failed [block != nullptr]: BasicBlock requested for unrecognized address
(BuilderBase.h:550 block_for_offset)
npm ERR! Lifecycle script `prisma:reset` failed with error:
npm ERR! Error: command failed
npm ERR!   in workspace: @klicker-uzh/prisma@3.0.0-alpha.82
npm ERR!   at location: /Users/julius/Desktop/IBF/klicker-uzh/packages/prisma
```

sjschlapbach (approved):
Looks great, results in a significant speed-up 😄 

For all Mac users with M1 / M2 chips that might have problems with the updated prisma version, this might be caused by a node installation, which runs on a x64 emulation (Rosetta):
- run `node -e 'console.log(process.arch)'` to print the version that is used
- If this prints `x64` on a Mac computer with Apple Silicon chip, this could result in problems
- Uninstall all node installations, install the latest v18 version again (installing through `nvm` works well)
- `node -e 'console.log(process.arch)'` should now print `arm64` and the issues with prisma should be resolved

PR #3667, opened 2023-06-10
style(apps/frontend-manage): improve styling of session components on session list with new collapsible

sjschlapbach:
This pull request uses an extended version of the Collapsible component to obtain a more compact session list element, making use of the new primary and secondary elements that can be passed to the design-system component (see [design-system commit](https://github.com/uzh-bf/design-system/commit/4533f529abe35a041c80d83c6e45ff68435c52f8)). This also ensures that the buttons are not below the extendible content / button, but at the same height.

BEFORE:
Screenshot 2023-06-09 at 19 02 14

AFTER:
Screenshot 2023-06-09 at 19 01 41


BShaq (commented):
It looks fine for me code-wise. However, the collapsible button is not centered when trying it out on my machine🤔

This is how it looks on Chrome:
image

And on Firefox:
image

sjschlapbach:
@BShaq Thanks for the review, this has just been identified as a bug in connecion with Tailwind CSS. Somehow the required styles are not properly exported from the `uzh-bf/design-system`.

sjschlapbach (commented):
Possible fixed that could be related to the observed styling problem in connection with components from the `uzh-bf/design-system`:
- https://stackoverflow.com/questions/69746121/using-nextjs-how-can-you-import-in-css-using-tailwind-css
- https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports

rschlaefli (approved):

PR #3766, opened 2023-07-15
enhance(docs): upgrade docusaurus and packages, prepare versioning for v2/v3, update release announcement

rschlaefli:
- latest deployed here https://klicker-uzh-docs.vercel.app/home
- upgraded to 3.0.0-alpha.0 to get support for react 18
- using the versioning dropdown, we can start working on v3 docs while being able to extend v2 docs as well (e.g., with a migration guide). in the fall, it will be critical to provide docs for both tools using versioning, as they run in parallel until the end of 2023.

wip: fixing docsearch


sjschlapbach (approved):
LGTM 👍 
The only thing I noticed regarding the styles is a weird double scrolling behavior within the roadmap iframe on the development page. However, this seems to be an issue that is not solvable on our end.

PR #3790, opened 2023-07-24
feat: authentication with next-auth and Switch EduID, migration to pnpm with improved monorepo workspace

rschlaefli:


apps/auth/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,129 @@
...
+              swissEduPersonUniqueID: { essential: true },
+            },
+          },
+          scope: 'openid email https://login.eduid.ch/authz/User.Read',
+        },
+      },
+      idToken: true,
+      checks: ['pkce', 'state'],
+
+      profile(profile) {
```
Comment (rschlaefli) [inline apps/auth/src/pages/api/auth/[...nextauth].ts]: - profile param and id/email are coming from EduID
- password, shortname, and lastLoginAt are generated and persisted to DB on first login (not overwritten on subsequential logins, but still regenerated and ignored on each login)

apps/auth/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,129 @@
...
+  token,
+  secret,
+}: {
+  token: string
+  secret: string
+}) {
+  return JWT.sign(token, secret)
+}
+
+function generateRandomString(length: number) {
```
Comment (rschlaefli) [inline apps/auth/src/pages/api/auth/[...nextauth].ts:22]: function to generate a random shortname for new users -> need functionality to change it on first login/welcome screen

apps/auth/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,129 @@
...
+import bcrypt from 'bcryptjs'
+import JWT from 'jsonwebtoken'
+import type { NextAuthOptions } from 'next-auth'
+import NextAuth from 'next-auth'
+import { DefaultJWT } from 'next-auth/jwt'
+import { Provider } from 'next-auth/providers/index'
+
+import prisma from 'src/lib/prisma'
+
+export async function decode({
```
Comment (rschlaefli) [inline apps/auth/src/pages/api/auth/[...nextauth].ts]: custom decode/encode functions -> we need these to ensure that JWT contents are not encrypted (which next-auth does by default), as that makes them unreadable from our graphql backend

apps/auth/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,129 @@
...
+  ],
+  session: {
+    strategy: 'jwt',
+  },
+  jwt: {
+    decode,
+    encode,
+  },
+  cookies: {
+    sessionToken: {
```
Comment (rschlaefli) [inline apps/auth/src/pages/api/auth/[...nextauth].ts:143]: ensure the session token is stored on the klicker.local (or klicker.uzh.ch) domains instead of auth.klicker.local -> makes them usable on all other apps on the common domain stem

apps/auth/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,129 @@
...
+  callbacks: {
+    async jwt({ token, user, account }) {
+      // if there are a user and account on the first invocation, we are logging in a user
+      // otherwise, the login is related to a participant
+      if (user && account) {
+        token.role = UserRole.USER
+      }
+      return token
+    },
+    async redirect({ url, baseUrl }) {
```
Comment (rschlaefli) [inline apps/auth/src/pages/api/auth/[...nextauth].ts:165]: need to override this, as the default redirect callback ensures that only redirects to the same domain are allowed (but we want to redirect from auth.klicker.local to manage.klicker.local etc)

apps/auth/src/pages/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Footer from 'shared-components/src/Footer'
+
+function SignInOutButton() {
+  const router = useRouter()
+
+  const { data: session } = useSession()
+
+  if (session) {
+    return (
+      <>
```
rschlaefli, line 14: this is not really necessary, as we will handle logout from the other apps, but does not hurt either

apps/auth/src/pages/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    )
+  }
+
+  return (
+    <Button
+      className={{ root: 'p-4' }}
+      onClick={() =>
+        signIn(process.env.NEXT_PUBLIC_EDUID_ID, {
+          callbackUrl:
+            (router.query?.redirectTo as string) ??
```
rschlaefli: allow to pass a redirectTo path in the URL -> this will need to be escaped when we want to use it (URL encoding), but this is just for first testing

packages/prisma/src/prisma/migrations/20230726182929_rename_session/migration.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "Session" RENAME TO "LiveSession";
```
rschlaefli, line 1: we need to rename the Session table to LiveSession as next-auth uses a Session table of its own -> first migration

packages/prisma/src/prisma/migrations/20230726183803_next_auth_users/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+  - You are about to drop the column `ssoType` on the `Account` table. All the data in the column will be lost.
+  - You are about to drop the column `ssoType` on the `ParticipantAccount` table. All the data in the column will be lost.
+  - A unique constraint covering the columns `[provider,providerAccountId]` on the table `Account` will be added. If there are existing duplicate values, this will fail.
+  - Added the required column `provider` to the `Account` table without a default value. This is not possible if the table is not empty.
+  - Added the required column `providerAccountId` to the `Account` table without a default value. This is not possible if the table is not empty.
+  - Added the required column `type` to the `Account` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DropIndex
+DROP INDEX "ParticipantAccount_ssoType_ssoId_key";
```
rschlaefli, line 15: create all necessary tables for next-auth (including a new Session table)

sjschlapbach (commented):
Login through EduID works well after the following changes:

1. To change the port forwarding to port 8088 instead of 8081, execute `sudo nano /etc/pf.anchors/ch.uzh.klicker` and change the corresponding ports.
2. Rerun `sudo pfctl -vnf /etc/pf.anchors/ch.uzh.klicker`
3. Rerun `sudo nano /etc/pf.conf`
4. Restart the entire application (including podman, etc.)
4.1. Visit the Klicker domains `pwa.klicker.local` or `manage.klicker.local`. If they show `Bad Gateway` errors, a reboot of the computer might solve the problem.
5. Go to `test.eduid.ch` and create a separate account for the testing environment of SWITCH. Note that the normal Edu-ID Accounts are separat from these. The "create account" button is placed right next to the login button on this page.
6. It should now be possible to visit `auth.klicker.local`, log in with the Edu-ID Test credentials and be forwarded to a new Klicker-Account.

sjschlapbach (approved):

PR #3795, opened 2023-08-11
 enhance(apps/frontend-manage): add error parsing to question creation and edit form

sjschlapbach:
Screenshot 2023-08-11 at 18 14 45

### DIFF
--- a/apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
+++ b/apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
@@ -997,7 +997,7 @@ function QuestionEditModal({
                               input: 'bg-white text-gray-500',
                             }}
                             placeholder={t('shared.generic.minLong')}
-                            data-cy="set-numerical-minimum"
+                            data={{ cy: 'set-numerical-minimum' }}
                             hideError
                           />
                           <div className="font-bold">
@@ -1010,7 +1010,7 @@ function QuestionEditModal({
                               input: 'bg-white text-gray-500',
                             }}
                             placeholder={t('shared.generic.maxLong')}
-                            data-cy="set-numerical-maximum"
+                            data={{ cy: 'set-numerical-maximum' }}
                             hideError
                           />
                         </div>
@@ -1035,7 +1035,7 @@ function QuestionEditModal({
                               input: 'bg-white text-gray-500',
                             }}
                             precision={0}
-                            data-cy="set-numerical-accuracy"
+                            data={{ cy: 'set-numerical-accuracy' }}
                             hideError
                           />
                         </div>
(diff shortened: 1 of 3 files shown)


apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -1096,13 +1127,16 @@ function QuestionEditModal({
...
+                        <FormikNumberField
                           name="options.restrictions.maxLength"
-                          type="number"
-                          className="mr-2 bg-opacity-50 border rounded w-44 border-uzh-grey-100 h-9 focus:border-primary-40"
+                          className={{
+                            field:
+                              'mr-2 bg-opacity-50 border rounded w-44  h-9 focus:border-primary-40',
+                          }}
                           placeholder={t('manage.questionForms.answerLength')}
-                          min={0}
```
rschlaefli, line 1104: why remove min? would make sense to be able to specify only positive integers here

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -968,26 +986,32 @@ function QuestionEditModal({
                   {questionType === QuestionType.Numerical && (
                     <div>
                       <div className="w-full">
-                        <div className="flex flex-row items-center gap-2">
+                        <div className="flex flex-row items-center gap-2 mb-2">
                           <div className="font-bold">
                             {t('shared.generic.min')}:{' '}
                           </div>
-                          <FastField
+                          <FormikNumberField
```
rschlaefli, line 993: is there any performance impact because of this change? we had used fast field to counteract the lagging in inputs. but certainly makes sense to migrate to our own fields for this

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
sjschlapbach: Because the current `FormikNumberField` component does not accept a minimum prop... This is still an open TODO, but only makes sense to be implemented after the validation issue on first render has been resolved before.

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
sjschlapbach: yes, there will probably be a small performance impact due to this change as the field will probably be validated more often. On the other hand, replacing it with our custom component results in a significantly more rigorous validation at the moment, wherefore the trade-off seemed to make sense.
The lagging was probably mainly caused by the editor component revalidating its content on every render, which seems like a minor problem here considering the length of the input?

rschlaefli (approved):

PR #3796, opened 2023-08-12
 enhance(apps/frontend-pwa): display student response scoring in learning element progress bar

sjschlapbach:
This pull request adds status highlighting to the learning element progress bar depending on the students answers:
- If all elements in a stack are answered correctly, the step will be highlighted as correct (green)
- If at least one element in a stack was only partially correct, the step will be highlighted as partially correct (orange)
- If all elements of a stack were answered incorrectly, the step will be highlighted as incorrect (red)
- Stacks with text elements only will be automatically highlighted as correct as soon as the user has navigated there

Screenshot 2023-08-12 at 17 43 50

### DIFF
--- a/apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
+++ b/apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
@@ -1,6 +1,6 @@
 import { IconDefinition } from '@fortawesome/free-regular-svg-icons'
 import {
-  faBookOpen,
+  faBarsStaggered,
   faCheck,
   faCheckDouble,
   faInbox,
@@ -73,7 +73,7 @@ function LearningElement({
                     <div>{element.score}p</div>
                   ) : (
                     <div>
-                      <FontAwesomeIcon icon={faBookOpen} />
+                      <FontAwesomeIcon icon={faBarsStaggered} />
                     </div>
                   )}
                   <FontAwesomeIcon icon={ICON_MAP[element.status]} />
--- a/apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
+++ b/apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
@@ -1,5 +1,6 @@
 import { LearningElement } from '@klicker-uzh/graphql/dist/ops'
 import { StepProgress } from '@uzh-bf/design-system'
+import { useState } from 'react'
 import { twMerge } from 'tailwind-merge'
 import ElementOverview from './ElementOverview'
 import ElementSummary from './ElementSummary'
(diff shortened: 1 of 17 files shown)


apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
```diff
@@ -19,6 +20,9 @@ function LearningElement({
   handleNextQuestion,
 }: LearningElementProps) {
   const currentStack = element.stacks?.[currentIx]
+  const [stepStatus, setStepStatus] = useState<
+    ('unanswered' | 'incorrect' | 'partial' | 'correct')[]
+  >(Array.from({ length: element.stacks?.length ?? 0 }, () => 'unanswered'))
```
rschlaefli: might be easier to read to use functions from ramda for such cases, e.g.,

```suggestion
  R.repeat('unanswered', element.stacks?.length ?? 0)
```

apps/frontend-pwa/src/components/learningElements/QuestionStack.tsx
```diff
@@ -143,16 +146,58 @@ function QuestionStack({
...
+          !element.questionInstance || element.questionInstance.evaluation
+      ) ??
+      false
+    ) {
+      setIsEvaluation(true)
+    }
+
+    let overallScore: 'correct' | 'incorrect' | 'partial' | 'unanswered' =
+      'unanswered'
+    stack.elements?.forEach((element) => {
```
rschlaefli: maybe generalize this logic and use reduce

packages/graphql/src/services/learningElements.ts
```diff
@@ -96,6 +97,7 @@ function evaluateQuestionResponse(
             pointsPercentage,
             multiplier: multiplier ?? 1,
           }),
+          percentile: pointsPercentage ?? 0,
```
rschlaefli, line 100: should we add these defaults in the grading package or does it make more sense here?

sjschlapbach: After checking the implementation again, I think it makes sense to leave it like this as 0 and null have different meanings in the grading package. Null is only returned, if no sample solutions are specified (i.e. content cannot be graded) and 0 is only returned when the answer was entirely wrong. Even though the case that there are no sample solutions cannot occur for learning elements (as we do not allow the use of such questions), it probably makes sense to still distinguish these cases in the grading package.

apps/frontend-pwa/src/components/learningElements/LearningElement.tsx
sjschlapbach: good idea, changed it

apps/frontend-pwa/src/components/learningElements/QuestionStack.tsx
sjschlapbach: the entire logic has been reworked into a single and more concise useEffect snipped with reduced complexity and better readability

rschlaefli (approved):

rschlaefli:
Extended version including points/scoring and icon for content-only parts (using custom formatter on StepProgress component)

MicrosoftTeams-image (1)

PR #3797, opened 2023-08-14
enhance: welcome page and optional participation

rschlaefli:


apps/backend-docker/src/app.ts
```diff
@@ -57,6 +57,10 @@ function prepareApp({ prisma, redisExec, pubSub, cache, emitter }: any) {
           // }
 
           if (req.cookies) {
+            if (req.cookies['no-account']) {
```
rschlaefli: if the account has been deleted, we set this cookie to ensure that no account is immediately recreated in, e.g., the OLAT environment (where we try to create an account on each page load). does not cover the case where you visit using a different browser or after cleaning cookies.

apps/frontend-control/src/lib/token.ts
```diff
@@ -1,98 +0,0 @@
-import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
```
rschlaefli, line 1: logic not needed in control (is used for LTI login in PWA)

apps/frontend-manage/src/pages/courses/[id].tsx
```diff
@@ -286,16 +286,6 @@ function CourseOverviewPage() {
                 points: course.averageActiveScore?.toFixed(2),
               })}
             </div>
-            <div className="mt-1">
```
Comment (rschlaefli) [inline apps/frontend-manage/src/pages/courses/[id].tsx:289]: as we dont want to collect the data for participants that do not participate on the leaderboard, they do not get an entry on the LB, and these numbers would be the same for both cases (total participants and active participants)

apps/frontend-pwa/src/components/common/Header.tsx
```diff
@@ -125,20 +125,29 @@ function Header({
...
             data={{ cy: 'header-avatar' }}
           >
             <Image
-              src={`${process.env.NEXT_PUBLIC_AVATAR_BASE_PATH}/${
-                participant?.avatar ?? 'placeholder'
-              }.svg`}
+              src={
+                participant?.avatar
+                  ? `${process.env.NEXT_PUBLIC_AVATAR_BASE_PATH}/${participant?.avatar}.svg`
+                  : '/user-solid.svg'
```
rschlaefli, line 142: new and more fitting default avatar

apps/frontend-pwa/src/components/learningElements/SingleQuestion.tsx
```diff
@@ -108,12 +108,6 @@ function SingleQuestion({
           <div className="flex flex-row items-center justify-between mb-4 border-b">
             <H3 className={{ root: 'mb-0' }}>{questionData.name}</H3>
             <div className="flex flex-row items-center gap-3">
-              <div className="text-sm md:text-base text-slate-500">
```
rschlaefli, line 111: removed this as it is quite buggy in stacked learning elements

apps/frontend-pwa/src/lib/token.ts
```diff
@@ -19,6 +16,11 @@ export async function getParticipantToken({
 
   const cookies = nookies.get(ctx)
 
+  // if the cookie no-account exists, return early
+  if (cookies?.['no-account'] === 'true') {
```
rschlaefli: skip trying to create a new account through LTI if the cookie has been set after deletion

apps/frontend-pwa/src/lib/token.ts
```diff
@@ -60,6 +62,7 @@ export async function getParticipantToken({
           variables: {
             courseId: query.courseId as string,
             participantId: request.body.lis_person_sourcedid,
+            email: request.body.lis_person_contact_email_primary,
```
rschlaefli: get the email address from LTI

apps/frontend-pwa/src/pages/course/[courseId]/docs/appSetup.tsx
```diff
@@ -7,6 +7,7 @@ function AppSetup() {
     <DocsLayout>
       {(courseInformation) => (
         <Markdown
+          withProse
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/course/[courseId]/docs/appSetup.tsx:10]: new option that uses <Prose> instead of <div> as a base element for the markdown component (which is a popular use case for markdown text)

apps/frontend-pwa/src/pages/course/[courseId]/docs/index.tsx
```diff
@@ -54,7 +54,7 @@ export async function getServerSideProps(ctx: GetServerSidePropsContext) {
     ctx,
   })
 
-  if (participant && !participant.avatar) {
+  if (participant && !participant.isActive) {
```
Comment (rschlaefli) [inline apps/frontend-pwa/src/pages/course/[courseId]/docs/index.tsx]: new field on participant that is set when the first save on the edit profile page occurs. this ensures that accounts are only active after the participant has seen this page at least once. reason: the page includes the data privacy disclaimer, which needs to be seen and acknowledged, before we can start tracking data. this also works for existing users, as we set the default to false, so they will see the page on first visit of the klicker in a new course.

apps/frontend-pwa/src/pages/editProfile.tsx
```diff
@@ -145,85 +164,179 @@ function EditProfile() {
             },
           })
 
-          if (result.data?.updateParticipantProfile) {
-            router.replace(decodedRedirectPath)
+          if (result.data?.updateParticipantProfile && !result.errors) {
+            // router.replace(decodedRedirectPath)
```
rschlaefli, line 168: no redirect after edit of profile, as there are multiple save buttons

apps/frontend-pwa/src/pages/editProfile.tsx
```diff
@@ -145,85 +164,179 @@ function EditProfile() {
...
+            <Form>
+              <div className="flex flex-col md:w-full md:max-w-5xl md:mx-auto gap-4">
+                <div className="flex flex-col w-full md:flex-row gap-4">
+                  <div className="flex flex-col justify-between flex-1 order-2 gap-3 md:order-1 md:bg-slate-50 md:p-4 rounded">
+                    <div>
+                      <H3 className={{ root: 'border-b mb-0' }}>
+                        {t('shared.generic.profile')}
+                      </H3>
+                      <div className="space-y-3 mb-2">
+                        <FormikTextField
```
rschlaefli, line 185: new field for email, still planning to disable this if LTI login is used, so that email cannot be edited for LTI accounts.

apps/frontend-pwa/src/pages/editProfile.tsx
```diff
@@ -145,85 +164,179 @@ function EditProfile() {
...
+                          className={{ label: 'font-bold text-md text-black' }}
+                          type="password"
+                        />
+                        <div>
+                          <div className="font-bold">
+                            {t('pwa.profile.publicProfile')}
+                          </div>
+                          <div className="flex flex-row space-between gap-4">
+                            <div className="flex flex-col items-center gap-1">
+                              <FormikSwitchField name="isProfilePublic" />
```
rschlaefli, line 218: new switch to toggle between self and others "Anonymous" and visibility of the pseudonyms

apps/frontend-pwa/src/pages/editProfile.tsx
```diff
@@ -145,85 +164,179 @@ function EditProfile() {
...
-                      />
-                    </div>
+                  <div className="flex-1 order-1 md:order-2 md:bg-slate-50 md:p-4 rounded justify-between flex flex-col space-y-4">
+                    {!data.self?.isActive && (
+                      <div className="flex-initial space-y-2">
+                        <H3 className={{ root: 'border-b mb-0' }}>Privacy</H3>
+                        <Markdown
+                          withProse
+                          withLinkButtons={false}
+                          content={t('pwa.profile.privacyNotice')}
```
rschlaefli, line 252: still being written, will be added to de.json and en.json once done, untranslated string for now

packages/graphql/src/services/accounts.ts
```diff
@@ -272,3 +272,23 @@ export async function changeParticipantLocale(
...
+  })
+
+  if (!participant) return false
+
+  await ctx.prisma.participant.delete({
+    where: { id: ctx.user.sub },
+  })
+
+  // set a cookie that lives forever and will be used to ensure that no new account is created through LTI
+  ctx.res.cookie('no-account', 'true', {
```
rschlaefli: set the no-account cookie to prevent account recreation in LTI setting

packages/graphql/src/services/courses.ts
```diff
@@ -119,6 +150,10 @@ export async function leaveCourse(
     },
   })
 
+  await ctx.prisma.leaderboardEntry.deleteMany({
```
rschlaefli, line 153: delete leaderboard entries corresponding to the current participation when a course is being left (means that all points from any activity are deleted)

packages/graphql/src/services/courses.ts
```diff
@@ -509,6 +548,11 @@ export async function getCourseData(
         orderBy: {
           score: 'desc',
         },
+        where: {
+          participation: {
+            isActive: true,
```
rschlaefli, line 555: only fetch active participations

packages/graphql/src/services/courses.ts
```diff
@@ -93,16 +99,41 @@ export async function joinCourse(
 
   if (!participation) return null
 
+  const lbEntry = await ctx.prisma.leaderboardEntry.create({
```
rschlaefli, line 102: create the leaderboard entry for the course as soon as participant joins the course. this prevents issues down the line and ensures we can simply update later on, instead of update or create on getting a question response anywhere.

packages/markdown/src/Markdown.tsx
```diff
@@ -107,30 +112,32 @@ function Markdown({
...
-                  >
-                    <div>
-                      {isExcel && <FontAwesomeIcon icon={faFileExcel} />}
-                      {isPDF && <FontAwesomeIcon icon={faFilePdf} />}
-                    </div>
-                    <div>{children}</div>
-                  </a>
-                )
-              },
+              a: withLinkButtons
```
rschlaefli, line 115: we do not want to parse links to buttons in all use cases (e.g., in the privacy notes), so this is a new prop to disable this functionality

packages/prisma/src/prisma/schema.prisma
```diff
@@ -735,12 +735,17 @@ model ParticipantAccount {
 model Participant {
   id String @id @default(uuid()) @db.Uuid
 
+  email          String? @unique
+  isEmailValid   Boolean @default(false)
```
rschlaefli, line 739: not yet used, as preparation for potential email validation for manual accounts (not LTI). for validated LTI, we could set to true automatically.

packages/shared-components/src/Leaderboard.tsx
```diff
@@ -81,59 +88,59 @@ function Leaderboard({
...
-              onParticipantClick
-                ? () => onParticipantClick(participant.id, true)
-                : undefined
-            }
-          />
+          <div className="max-w-none p-2 bg-slate-100 rounded border-slate-300 border text-slate-600 text-sm">
+            <Markdown
+              withProse
+              withLinkButtons={false}
+              content={t('pwa.general.joinLeaderboardNotice')}
```
rschlaefli, line 137: still being written, will be added to de.json and en.json once done, untranslated string for now

apps/frontend-manage/src/pages/courses/[id].tsx
Comment (sjschlapbach) [reply apps/frontend-manage/src/pages/courses/[id].tsx:289]: makes sense - also removed the corresponding unused strings from the language translation files

apps/frontend-pwa/src/components/learningElements/SingleQuestion.tsx
sjschlapbach: It is not really buggy, just hasn't been fixed since the switch from single questions to stacks :) / still an open TODO

apps/frontend-pwa/src/pages/course/[courseId]/index.tsx
```diff
@@ -269,9 +270,7 @@ function CourseOverview({ courseId }: Props) {
               {/* // TODO: update the translation strings as well, once this hard-coded content has been updated with a flexible implementation */}
               {course?.awards && course?.awards?.length != 0 && (
                 <div className="px-4 py-3 mt-4 bg-orange-100 border border-orange-200 rounded shadow md:mt-6">
-                  <H3 className={{ root: 'mb-2 text-base' }}>
-                    BF-Champion Awards
-                  </H3>
+                  <H3 className={{ root: 'mb-2 text-base' }}>Awards</H3>
```
Comment (sjschlapbach) [inline apps/frontend-pwa/src/pages/course/[courseId]/index.tsx:273]: We should probably also translate the "Awards" string to also have a consistent German page?

apps/frontend-pwa/src/pages/editProfile.tsx
```diff
@@ -145,85 +164,179 @@ function EditProfile() {
...
+                        skinTone={values.skinTone}
+                        clothingColor={values.clothingColor}
+                        eyes={values.eyes}
+                        hair={values.hair}
+                        hairColor={values.hairColor}
+                        mouth={values.mouth}
+                        facialHair={values.facialHair}
+                      />
+                    </div>
+                    <div className="grid grid-cols-2 md:grid-cols-3 gap-2 text-xs">
```
sjschlapbach: Text appears to be very small on larger laptop screens / desktop screens. Maybe we could scale up the font a bit on larger screens?

sjschlapbach:
```suggestion
                    <div className="grid grid-cols-2 md:grid-cols-3 gap-2 text-xs md:text-sm">
```

packages/shared-components/src/Leaderboard.tsx
```diff
@@ -81,59 +88,59 @@ function Leaderboard({
...
+            />
+            <Button fluid className={{ root: 'bg-white' }} onClick={onJoin}>
+              Join the leaderboard for{' '}
```
sjschlapbach, line 140: this will also need a translation - can be combined with the join notice right above

packages/shared-components/src/constants.ts
```diff
@@ -149,3 +149,529 @@ export const PRESET_COURSE_COLORS = [
   '#FE7408',
   '#D84B39',
 ]
+
+export const ANIMAL_NAMES = [
```
sjschlapbach, line 153: What is the purpose of those? They seem not to be in use at the moment?

apps/backend-docker/src/app.ts
rschlaefli: removed because of change in account creation approach with dedicated page

packages/shared-components/src/constants.ts
rschlaefli: maybe later :)

PR #3803, opened 2023-08-20
enhance: finalize integration of privacy policy and ToS

rschlaefli:


apps/auth/src/pages/index.tsx
```diff
@@ -20,9 +24,36 @@ function SignInOutButton() {
 
   return (
     <div className="flex flex-col gap-4">
+      <Checkbox
```
rschlaefli, line 65: auth frontend is EN only for now. not sure it is worth it to translate this at all..

apps/auth/src/pages/index.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { useRouter } from 'next/router'
 
+import useStickyState from 'src/hooks/useStickyState'
+
 function SignInOutButton() {
   const router = useRouter()
 
   const { data: session } = useSession()
 
+  const [tosChecked, setTosChecked] = useStickyState('tos-agreement', false)
```
rschlaefli: use sticky state -> ensure the ToS do not have to be accepted on every single login, but only once per browser (store accept in local storage)

apps/docs/package.json
```diff
@@ -39,13 +39,13 @@
   "scripts": {
     "build": "docusaurus build",
     "deploy": "docusaurus deploy",
+    "dev:docs": "docusaurus start  --port 5000",
```
rschlaefli, line 43: pn dev:docs on top level starts the docs. dev does not include them otherwise.

apps/frontend-control/src/pages/course/[id].tsx
```diff
@@ -75,7 +75,7 @@ function Course() {
 export async function getStaticProps({ locale }: GetStaticPropsContext) {
   return {
     props: {
-      messages: (await import(`@klicker-uzh/i18n/messages/${locale}.json`))
+      messages: (await import(`@klicker-uzh/i18n/messages/${locale}`))
```
Comment (rschlaefli) [inline apps/frontend-control/src/pages/course/[id].tsx:78]: all of these change because we now import from a .ts file

apps/auth/src/pages/index.tsx
sjschlapbach: would probably make sense as it is legal stuff, that should be understood rigorously? Also, it is not too much to translate compared to the other frontends, I can look into this.

sjschlapbach (commented):
LGTM

PR #3818, opened 2023-08-27
feat: image uploads and media library

rschlaefli:


apps/frontend-manage/src/components/common/ContentInput.tsx
rschlaefli, line 1: moved from shared components to frontend manage, as it now also includes graphql requests and more dependencies (azure blob, dropzone)

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -83,18 +83,6 @@ function QuestionEditModal({
     hasSampleSolution: Yup.boolean(),
     hasAnswerFeedbacks: Yup.boolean(),
 
-    // TODO: adapt validation structure for attachments once they are available
-    attachments: Yup.array()
```
rschlaefli, line 84: completely removed attachments from all code

packages/graphql/src/schema/query.ts
```diff
@@ -105,8 +105,30 @@ export const Query = builder.queryType({
       userTags: asUser.prismaField({
         nullable: true,
         type: [Tag],
-        resolve(_, __, ___, ctx) {
-          return QuestionService.getUserTags(ctx)
+        async resolve(_, __, ___, ctx) {
```
rschlaefli, line 108: moved some of the very simple services here. when we split up the long query and mutation files anyways, we might as well merge some of the simple services, so we do not have it distributed across so many files.

packages/graphql/src/services/questions.ts
```diff
@@ -363,3 +343,74 @@ export async function toggleIsArchived(
...
+    process.env.BLOB_STORAGE_ACCESS_KEY as string
+  )
+
+  const storageAccount = `https://${
+    process.env.BLOB_STORAGE_ACCOUNT_NAME as string
+  }.blob.core.windows.net`
+
+  // if nonexistent, create a container for the user on blob storage
+  const client = new BlobServiceClient(storageAccount, sharedKeyCredential)
+  const containerClient = client.getContainerClient(ctx.user.sub)
```
rschlaefli, line 373: each user gets a container in blob storage with the user id. the generated token allows for a file upload to exactly one location in this container as specified by the randomly generated UUID.

rschlaefli: the migration will put all of the existing images into this container as well and adds them to the media library.

apps/frontend-manage/src/components/common/ContentInput.tsx
sjschlapbach: Would it make sense to still keep it in the shared components and somehow extend it within a frontend-manage component (e.g. thinking about rich text answers for students in frontend-pwa?). However, as these rich text answers are not yet available anyways, we can also postpone this.

packages/prisma/src/prisma/migrations/20230827124621_media_files/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+
+*/
+-- DropForeignKey
+ALTER TABLE "Attachment" DROP CONSTRAINT "Attachment_ownerId_fkey";
+
+-- DropForeignKey
+ALTER TABLE "Attachment" DROP CONSTRAINT "Attachment_questionId_fkey";
+
+-- DropForeignKey
+ALTER TABLE "AttachmentInstance" DROP CONSTRAINT "AttachmentInstance_questionInstanceId_fkey";
```
sjschlapbach, line 15: These modifications do not really make sense, if we drop the `Attachment` and `Attachment tables afterwards anyways, right?`

apps/frontend-manage/src/components/common/ContentInput.tsx
rschlaefli: probably, but we need to rethink the shared component approach or have multiple shared component packages (e.g., one just for react components, etc., much easier now). would be useful to have the content input on frontend-pwa for formatted student inputs in the future, but without the need to install some backend dependencies from peerDeps, etc.

packages/prisma/src/data/data/TEST.ts
```diff
@@ -1,21 +1,6 @@
...
-    type: AttachmentType.SVG,
-  },
-  {
-    id: 'b0b9c0c0-0b0b-4b4b-0b0b-0b0b0b0b0b0c',
-    href: 'https://sos-ch-dk-2.exo.io/klicker-uzh-dev/avatars/41b98856a8c221db667cf066f34b931eff048c32.webp',
-    name: 'Test Attachment 2',
-    type: AttachmentType.WEBP,
-  },
-]
+const { QuestionType, SessionStatus, OrderType } = Prisma
```
sjschlapbach, line 3: Do we still have some example data in one of the buckets, which we could include in a seed so that images are not just left out and might break during development / layout refactors, when not checked manually and explicitely?

apps/frontend-manage/src/components/common/ContentInput.tsx
sjschlapbach: ah yes, that makes total sense 👍

packages/prisma/src/prisma/migrations/20230827124621_media_files/migration.sql
rschlaefli: these files are generated, I think you need to drop constraints before you can delete a postgres table, but not sure :) just what prisma does.

packages/prisma/src/data/data/TEST.ts
rschlaefli: TBD later together with a cypress test that uploads some files (but will need to clean them from azure afterwards, so probably also some deletion mechanism required)

PR #3819, opened 2023-08-27
enhance(migration): migration to media files, add templates and fix email sending

rschlaefli:


apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -454,7 +454,7 @@ function QuestionEditModal({
 
         return (
           <Modal
-            // asPortal
+            asPortal
```
rschlaefli, line 457: found out that this is necessary to ensure content inputs work

PR #3842, opened 2023-09-08
enhance(apps/frontend-manage): ensure that there are exactly four answers for kprim questions

BShaq:


apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -929,8 +936,17 @@ function QuestionEditModal({
...
                             className={{
-                              root: 'font-bold border border-solid border-uzh-grey-100',
+                              root: twMerge(
+                                'font-bold border border-solid border-uzh-grey-100',
+                                values.type === QuestionType.Kprim &&
+                                  values.options.choices.length >= 4 &&
+                                  'opacity-50 cursor-not-allowed'
+                              ),
                             }}
+                            disabled={
```
sjschlapbach: Should we maybe directly show 4 empty answer options for KPRIM questions as we do not allow less anyways at the moment? Then we could hide the "add answer button" completely for this question type and maybe even simplify the validation?

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
```diff
@@ -140,6 +137,16 @@ function QuestionEditModal({
             })
           }
 
+          case 'KPRIM': {
+            const choicesSchema = baseChoicesSchema
+              .min(4, t('manage.formErrors.NumberQuestionsRequiredKPRIM'))
```
sjschlapbach: I found on StackExchange that cypress might have an option "should('have.length', 4)" to check for an exactly length?

sjschlapbach (commented):
Looks great and should solve a lot of confusion that might occur with these question types 😅 Maybe we would also have to adapt the created questions in the test suite to ensure passing tests again? (I did not check the corresponding points of failure though)

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
BShaq: I understand your suggestion about directly displaying four empty answer options for KPRIM questions, especially since we don't allow fewer than that. It seems like a good way to enhance user experience by simplifying the interface.

However, implementing this change is proving to be more complex than initially thought:

1. Existing Bug: During the implementation, I discovered an existing bug, which gets triggered when:

- A user clicks on Add new answer.
- Then, deletes the top-most field.
- Upon attempting to modify the content of the newly added field, it inadvertently creates a new field on the click action. This behavior isn't new as I could verify it on the V3 branch.

2. The above behavior complicates KPRIM question management. Consider this scenario:

- A user creates 5 answer options for an SC question.
- They switch the question type to KPRIM. Now, all 5 answers are displayed, which goes against the KPRIM format.
- The user therefore needs to delete one answer as only 4 are allowed. However, if they try to remove the 4th field and then change the content of the last one (which was previously the 5th field), the system again creates a new field on click, even duplicating the content if present.

3. Preview Discrepancies: This issue has also caused discrepancies in the number of fields displayed in the preview. Addressing this is also planned for this PR.

Anyways, I'm currently working on addressing all of these issues :)

apps/frontend-manage/src/components/questions/QuestionEditModal.tsx
BShaq: thanks for pointing that out. But it seems that `should()` does not work for Yup's ArraySchema anymore. Instead, I used now `length()` to specify the desired length

rschlaefli (approved):

PR #3870, opened 2023-09-25
 feat: add option to apply question changes to all planned elements - live session, microlearning, practice quiz

sjschlapbach:


packages/graphql/src/graphql/ops/MUpdateQuestionInstances.graphql
```diff
@@ -0,0 +1,5 @@
+mutation UpdateQuestionInstances($questionId: Int!) {
+  updateQuestionInstances(questionId: $questionId) {
+    id
```
sjschlapbach, line 3: @rschlaefli Should we return the entire instances here to update the cache properly? I left it out for now as the instances should hopefully be fetched together with the corresponding element anyways?

sjschlapbach (commented):
Once this is merged, this should also be reflected in comments on the following topics:
https://klicker-uzh.feedbear.com/boards/feature-requests/posts/questions-update-changed-questions-in-planned-session-and-activities
https://community.klicker.uzh.ch/t/update-of-question-in-sessions/117

packages/graphql/src/services/questions.ts
```diff
@@ -415,3 +416,112 @@ export async function getFileUploadSas(
...
+      id: questionId,
+    },
+    include: {
+      instances: {
+        include: {
+          sessionBlock: {
+            include: {
+              session: true,
+            },
+          },
```
BShaq, line 436: I think we still need to filter for the relevant sessions here, i.e., with status PREPARED or SCHEDULED. Otherwise the questions will be updated also in running sessions, too.

packages/i18n/messages/de.ts
```diff
@@ -746,6 +746,9 @@ Da die KlickerUZH-App noch nicht im iOS-App-Store verfügbar ist, folgen Sie die
       uploadImageHeader: 'Medien hinzufügen',
       uploadImageDescription:
         'Ziehen Sie ein Bild auf diese Fläche oder klicken Sie darauf, um den Explorer zu öffnen.',
+      updateInstances: 'Frage-Instanzen in KlickerUZH-Elementen aktualisieren',
+      updateInstancesExplanation:
+        'Nutzen sie diese Einstellung, um die Frage in allen geplanten Sessionen, Lernelementen und Micro-Sessions anzupassen. Der Inhalt von Fragen in laufenden und abgeschlossenen Elementen wird nicht aktualisiert. Veränderte Multiplier werden auch auf die erstellten Instanzen angewendet.',
```
BShaq:
```suggestion
        'Nutzen sie diese Einstellung, um die Frage in allen geplanten Quizzes, Übungs-Quizzes und Microlearnings anzupassen. Der Inhalt von Fragen in laufenden und abgeschlossenen Elementen wird nicht aktualisiert. Veränderte Multiplier werden auch auf die erstellten Instanzen angewendet.',
```

packages/i18n/messages/en.ts
```diff
@@ -749,6 +749,9 @@ Since the KlickerUZH app is not yet available on the iOS App Store, follow these
       uploadImageHeader: 'Upload Media',
       uploadImageDescription:
         'Drag an image here to upload or click to open file explorer.',
+      updateInstances: 'Update question instances in KlickerUZH elements',
+      updateInstancesExplanation:
+        'Use this setting to update the question in all planned sessions, learning elements and micro-sessions. The content of questions in running and completed elements will not be updated. Changed multipliers will also be applied to the created instances.',
```
BShaq:
```suggestion
        'Use this setting to update the question in all planned quizzes, practice quizzes and microlearnings. The content of questions in running and completed elements will not be updated. Changed multipliers will also be applied to the created instances.',
```

packages/i18n/messages/de.ts
BShaq: The naming of those elements has changed 😅

packages/graphql/src/services/questions.ts
```diff
@@ -415,3 +416,112 @@ export async function getFileUploadSas(
...
+    },
+  })
+
+  if (!question) {
+    return null
+  }
+
+  // get all instances and the corresponding element multipliers
+  const instanceMultipliers: Record<number, number> = {
+    ...question.instances.reduce((prev1, instance) => {
```
BShaq: A bit nitpicking, but I think `prev1`  is not a very intuitive variable name.. What would you think of renaming it to something like `acc` referring to the accumulator which gathers all the results from previous iterations?🤔

BShaq (changes requested):
I highlighted some minor points. Once these are addressed, the PR should be ready :)

packages/i18n/messages/de.ts
sjschlapbach: good catch, thanks!

packages/graphql/src/services/questions.ts
sjschlapbach: Yes, it would be cool to have this here, but Prisma does not allow to do it for some reason. Only "include" and "data" can be entered here "where" is not accepted and causes the query to crash. However, as the corresponding sessions are removed anyway further down, it should still be fine :)

sjschlapbach:
Changes to v3.1-dev base on PR #3941 with changes for invalidation of the parent elements

PR #3871, opened 2023-09-26
enhance(packages/grading): update grading logic and test suite

sjschlapbach:
This pull request contains the following updates in the grading package:
- The default points, which are awarded for participating in a live session should not be multiplied by either the question multiplier nor the element multiplier, as wrong answers can lead to a lot of points otherwise
- The test suite is extended with different point combination functions
- The function to compute experience points is updated not to take a multiplier anymore, as it is not considered afterwards anyways
- The grading test suite is now also run as a part of the CI actions

### DIFF
--- a/.github/workflows/test-grading.yml
+++ b/.github/workflows/test-grading.yml
@@ -22,5 +22,4 @@ jobs:
           run_install: true
       - name: Test functions
         shell: bash
-        run: |
-          pnpm run test
+        run: jest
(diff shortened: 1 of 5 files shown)


sjschlapbach (commented):
Once this is merged, we have to upgrade the logic as explained in the live session paragraph in the community use case (plots are fine as they do not contain any multpliers in any case):
https://community.klicker.uzh.ch/t/use-case-grading-approach-for-gamified-interactions/132

rschlaefli (approved):

PR #3875, opened 2023-09-27
enhance: add teams notifications for critical actions

sjschlapbach:


apps/auth/src/lib/util.ts
```diff
@@ -0,0 +1,15 @@
+import axios from 'axios'
+
+export async function sendTeamsNotifications(scope: string, text: string) {
+  if (process.env.TEAMS_WEBHOOK_URL) {
+    return axios.post(process.env.TEAMS_WEBHOOK_URL, {
+      '@context': 'https://schema.org/extensions',
+      '@type': 'MessageCard',
+      themeColor: '0076D7',
+      title: `Migration: ${scope}`,
```
rschlaefli:
```suggestion
      title: scope,
```

packages/graphql/src/lib/util.ts
```diff
@@ -38,4 +39,18 @@ export function formatDate(dateTime: Date) {
   }
 }
 
+export async function sendTeamsNotifications(scope: string, text: string) {
+  if (process.env.TEAMS_WEBHOOK_URL) {
+    return axios.post(process.env.TEAMS_WEBHOOK_URL, {
+      '@context': 'https://schema.org/extensions',
+      '@type': 'MessageCard',
+      themeColor: '0076D7',
+      title: `Migration: ${scope}`,
```
rschlaefli:
```suggestion
      title: scope,
```

packages/graphql/src/services/sessions.ts
```diff
@@ -879,171 +910,176 @@ export async function deactivateSessionBlock(
...
-    ctx,
-    sessionId,
-    sessionBlockId,
-    activeInstanceIds,
-  })
+    return updatedSession
+  } catch (error) {
+    await sendTeamsNotifications(
+      'graphql/deactivateSessionBlock',
+      `ERROR - failed to deactivate session block ${sessionBlockId} in session ${session.id} with active block ${session.activeBlockId}: ${error}`
```
rschlaefli: @sjschlapbach tested that this outputs something, i.e., error is not an object that cannot be printed? or should we use error.message or similar?

rschlaefli (changes requested):

packages/graphql/src/services/sessions.ts
sjschlapbach: Using error.message would probably be a good idea, will be fixed

rschlaefli (approved):

PR #3877, opened 2023-09-27
fix(packages/graphql): ensure that time on push notifications is fixed to CET

sjschlapbach:


packages/graphql/src/lib/util.ts
```diff
@@ -31,11 +33,11 @@ export function checkCronToken(ctx: Context) {
 }
 
 export function formatDate(dateTime: Date) {
-  let date = dayjs(dateTime).local()
+  let date = dayjs(dateTime).utc().tz('Europe/Zurich')
```
rschlaefli:
```suggestion
  let date = dayjs(dateTime).utc().tz(process.env.TZ ?? 'Europe/Zurich')
```

rschlaefli (approved):

PR #3882, opened 2023-09-28
refactor: add types for JSON fields and improve generic type structure

rschlaefli:


packages/graphql/src/schema/query.ts
```diff
@@ -361,7 +361,7 @@ export const Query = builder.queryType({
           id: t.arg.string({ required: true }),
         },
         resolve(_, args, ctx) {
-          return SessionService.getSessionEvaluation(args, ctx)
+          return SessionService.getSessionEvaluation(args, ctx) as any
```
rschlaefli, line 364: remove as any once all type issues in sessions service are fixed

packages/graphql/src/schema/question.ts
```diff
@@ -79,6 +79,9 @@ export const OptionsFreeTextInput = builder.inputType('OptionsFreeTextInput', {
 })
 
 export const ResponseInput = builder.inputType('ResponseInput', {
+  directives: {
```
rschlaefli: this together with the directives plugin will allow to ensure that exactly one type of response is sent (not empty and not free and choice response together)

packages/graphql/src/services/accounts.ts
```diff
@@ -783,14 +784,14 @@ async function seedDemoQuestions(ctx: ContextWithUser) {
         create: blockData.map(
           ({ questions, randomSelection, timeLimit }, blockIx) => {
             const newInstances = questions.map((question, ix) => {
-              const processedQuestionData = processQuestionData(question)
```
rschlaefli, line 786: processQuestionData has been removed, as it is a no-op in the current structure

packages/graphql/src/services/sessions.ts
```diff
@@ -17,38 +18,19 @@ import { GraphQLError } from 'graphql'
...
+  AllQuestionInstanceTypeData,
+  AllQuestionTypeData,
+  QuestionResults,
+  QuestionResultsChoices,
+} from 'src/types/app'
 import { sendTeamsNotifications } from '../lib/util'
 
 const scheduledJobs: Record<string, any> = {}
 
-export function processQuestionData(question: Question): AllQuestionTypeData {
```
rschlaefli, line 24: removed, as it was only used because of manual typing. if we want to "pick" or do anything else, this would need to be reintroduced in a betterway.

packages/graphql/src/services/sessions.ts
```diff
@@ -1638,11 +1623,15 @@ export async function getPinnedFeedbacks(
   return reducedSession
 }
 
-function checkCorrectnessFreeText(instance) {
+function checkCorrectnessFreeText(instance: AllQuestionInstanceTypeData) {
   // Adds "correct" attribute (true/false) to results in FREE_TEXT questions if they match any given solution)(exact match, case insensitive)
-  if (instance.questionData.type === 'FREE_TEXT') {
+  instance.elementType = instance.questionData.type
+  if (
+    instance.elementType === 'FREE_TEXT' &&
```
rschlaefli, line 1603: this is a very ugly trial to get results and questionData type narrowing to work within the same function. we need to find a better way for this, maybe add the elementType to the data schema of the instance in addition to the type in questionData, and use instance.elementType for all checks, also to detect the type of questionData.

packages/graphql/src/services/sessions.ts
```diff
@@ -1659,14 +1648,18 @@ function checkCorrectnessFreeText(instance) {
   return instance
 }
 
-function computeStatistics(instance) {
+function computeStatistics(instance: AllQuestionInstanceTypeData) {
   // Compute the statistics for numerical questions
-  if (instance.questionData.type === 'NUMERICAL' && !instance.statistics) {
+  instance.elementType = instance.questionData.type
+  if (
+    instance.elementType === 'NUMERICAL' &&
```
rschlaefli, line 1628: same as above

packages/graphql/src/services/sessions.ts
```diff
@@ -1703,20 +1703,22 @@ function computeStatistics(instance) {
...
-    }
+    instance.statistics = hasResults
+      ? {
+          max: max(valueArray),
+          mean: mean(valueArray),
+          median: median(valueArray),
+          min: min(valueArray),
+          q1: quantileSeq(valueArray, 0.25) as number,
+          q3: quantileSeq(valueArray, 0.75) as number,
+          sd: std(valueArray) as number[],
```
rschlaefli, line 1687: no idea why this is a number[], checking this would be good

packages/graphql/src/types/app.ts
```diff
@@ -0,0 +1,158 @@
+import {
```
rschlaefli, line 1: this file is new and defines all of the custom types for backend usage. the types as added to PrismaJson at the end can be used within the prisma schema as part of the special comments syntax.

packages/prisma/src/prisma/schema.prisma
```diff
@@ -178,13 +178,15 @@ model Question {
...
   name               String
   content            String
-  options            Json
   explanation        String?
   pointsMultiplier   Int                 @default(1)
   displayMode        QuestionDisplayMode @default(LIST)
   hasSampleSolution  Boolean             @default(false)
   hasAnswerFeedbacks Boolean             @default(false)
 
+  /// [PrismaQuestionOptions]
```
rschlaefli, line 187: this syntax will cause the json addon for prisma to replace all generic Json types with the type stated here. the addon works purely on the typescript level and output and does not change any logic.

rschlaefli (commented):
@sjschlapbach @BShaq initial draft seems to be working, but types in the backend code still need a cleanup, as there are many issues left, (also new ones because of the new typings)

packages/graphql/src/schema/query.ts
rschlaefli: also holds for all of the other as any in query.ts or mutation.ts, once we have clean typings everywhere, we should be able to remove all of them.

BShaq (dismissed):
LGTM :)

EDIT: wrong branch approved, sorry for that  😆

packages/graphql/src/services/sessions.ts
sjschlapbach: Should we maybe re-introduce this directly with this branch in an improved fashion so that we do not need to search later on where the processQuestionData has been used previously and possibly miss an occurrence?

packages/graphql/src/types/app.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+export type QuestionOptions =
+  | QuestionOptionsChoices
+  | QuestionOptionsNumerical
+  | QuestionOptionsFreeText
+
+interface IQuestionData<
+  Type extends QuestionType,
+  Options extends QuestionOptions
```
sjschlapbach: Why do we need to extend here and cannot directly use the QuestionType or QuestionOptions as generic types? Doesn't this work for technical reasons or do we actually want to allow more attributes than specified in those types?

sjschlapbach (changes requested):
In the microLearning.ts service, line 309, the function processQuestionData is still used. As we will use this again later anyway, should we remove it here and then re-introduce it asap or should we maybe directly correct the functionality of this function instead of removing it?

packages/graphql/src/types/app.ts
rschlaefli: this tells TS that the generic type corresponds to QuestionType, but we use a different name for the generic as there could be naming conflicts otherwise (when we, e.g., use QuestionType in the type: Type of the interface, it could mean either the generic or the prisma type.

sjschlapbach: that makes sense, thanks for the explanation :)

packages/graphql/src/services/learningElements.ts
```diff
@@ -78,9 +79,9 @@ function evaluateQuestionResponse(
           }),
           percentile: pointsPercentage ?? 0,
         }
-      } else if (data.type === QuestionType.MC) {
+      } else if (questionData.type === QuestionType.MC) {
         const pointsPercentage = gradeQuestionMC({
-          responseCount: data.options.choices.length,
+          responseCount: questionData.options.choices.length,
           response: response.choices!,
```
sjschlapbach: Is there a specific reason, why we still need to force TS to accept `response.choices` here as it seems to accept it for SC questions above? - same for KPRIM below

packages/graphql/src/services/learningElements.ts
```diff
@@ -194,161 +193,155 @@ export async function respondToQuestionInstance(
...
+            response.value === ''
+          ) {
+            return {}
+          }
 
-        const value = String(parsedValue)
+          const parsedValue = parseFloat(response.value)
+          if (
+            isNaN(parsedValue) ||
+            (typeof questionData.options.restrictions!.min === 'number' &&
```
sjschlapbach: TS correctly recognizes the questionData.options object here as `QuestionOptionsNumerical`. The restrictions are optional on this type right now, but still it is assumed that they exist. Should we either assume that the restriction could possibly not exist as suggested by the type or enforce on the type that numerical options have restrictions, even though min and/or max on them might not be defined?

packages/graphql/src/services/learningElements.ts
```diff
@@ -194,161 +193,155 @@ export async function respondToQuestionInstance(
...
-              questionData.options.restrictions!.maxLength)
-        ) {
-          return {}
-        }
+        case QuestionType.FREE_TEXT: {
+          if (
+            typeof response.value === 'undefined' ||
+            response.value === null ||
+            response.value === '' ||
+            (typeof questionData.options.restrictions!.maxLength === 'number' &&
```
sjschlapbach: Same here as above for numerical question type as TS again correctly recognizes `QuestionOptionsFreeText`

packages/graphql/src/lib/questions.ts
```diff
@@ -0,0 +1,65 @@
...
+  AllQuestionTypeData,
+  ChoicesQuestionData,
+  FreeTextQuestionData,
+  NumericalQuestionData,
+  QuestionResults,
+  QuestionResultsChoices,
+} from 'src/types/app'
+
+export function processQuestionData(question: Question) {
+  const extractRelevantKeys = R.pick([
```
sjschlapbach: Could we extract these keys from a GraphQL type for questionData for example using Object.keys(xy) in order to avoid hard-coded parameters?

packages/graphql/src/services/learningElements.ts
rschlaefli: changed to type casting. type narrowing does not work here, as response is on the instance, not within question data. maybe would work if we would narrow types on the question instance, but that will require prisma schema changes.

packages/graphql/src/services/learningElements.ts
rschlaefli: changed to ?

packages/graphql/src/services/learningElements.ts
rschlaefli: true, fixed with ?

sjschlapbach: types should be consistent then, yes :)

packages/graphql/src/lib/questions.ts
rschlaefli: good idea & done :)

sjschlapbach (approved):

PR #3883, opened 2023-09-28
enhance(apps/frontend-manage): allow to preset live session settings

sjschlapbach:
Screenshot 2023-09-28 at 23 37 37


apps/frontend-manage/src/components/sessions/creation/LiveSessionWizard.tsx
```diff
@@ -376,71 +386,124 @@ function StepTwo(props: StepProps) {
...
+            data={{ cy: 'set-gamification' }}
+            className={{ tooltip: 'z-20' }}
+          />
+          <ErrorMessage
+            name="isLiveQAEnabled"
+            component="div"
+            className="text-sm text-red-400"
+          />
+        </div>
+        {values.isLiveQAEnabled && (
```
BShaq: What do you think of simply disabling this functionality in case Live Q&A is disabled? This way, professors can directly see what settings can be preset and this behaviour would also be more consistent (e.g., Gamification & Multiplier in Session Creation and  Live Q&A and Moderation in Lecturer Cockpit)

BShaq (commented):
One minor thing, otherwise LGTM :)

apps/frontend-manage/src/components/sessions/creation/LiveSessionWizard.tsx
sjschlapbach: You mean that the setting cannot be changed if the live QA is not enabled? This should already be implemented, if I remember it correctly. That the default value for this is set to true, even if the live QA is disabled, was a deliberate decision, which was taken quite some time ago :)

BShaq: Sorry, I was not that clear😅 What I really meant was instead of hiding the moderation switch if Live QA is not enabled, we can simply disable the switch as was done the same in the mentioned examples🤔

sjschlapbach: Ah right, sorry, I thought it would be smart to hide it for disabled live QA as the default option of the moderation is true and the lecturer cannot disable it then initially...

sjschlapbach: Implementation has been adapted - switches are always shown, disabling is highlighted more explicitly in new design-system version, thanks for the feedback :)

BShaq (approved):

apps/frontend-manage/src/components/sessions/creation/LiveSessionWizard.tsx
```diff
@@ -376,71 +386,122 @@ function StepTwo(props: StepProps) {
...
+          />
+        </div>
+        <div>
+          <FormikSwitchField
+            disabled={!values.isLiveQAEnabled}
+            name="isModerationEnabled"
+            label={t('shared.generic.moderation')}
+            tooltip={t('manage.sessionForms.liveSessionModeration')}
+            standardLabel
+            data={{ cy: 'set-gamification' }}
```
rschlaefli:
```suggestion
            data={{ cy: 'set-liveqa-moderation' }}
```

apps/frontend-manage/src/components/sessions/creation/LiveSessionWizard.tsx
```diff
@@ -376,71 +386,122 @@ function StepTwo(props: StepProps) {
...
+            className="text-sm text-red-400"
+          />
+        </div>
+        <div>
+          <FormikSwitchField
+            name="isLiveQAEnabled"
+            label={t('shared.generic.liveQA')}
+            tooltip={t('manage.sessionForms.liveSessionLiveQA')}
+            standardLabel
+            data={{ cy: 'set-gamification' }}
```
rschlaefli:
```suggestion
            data={{ cy: 'set-liveqa-enabled' }}
```

apps/frontend-manage/src/components/sessions/creation/LiveSessionWizard.tsx
```diff
@@ -376,71 +386,122 @@ function StepTwo(props: StepProps) {
...
+        </div>
+      )}
+      <div className="flex flex-col gap-4">
+        <div>
+          <FormikSwitchField
+            name="isConfusionFeedbackEnabled"
+            label={t('shared.generic.feedbackChannel')}
+            tooltip={t('manage.sessionForms.liveSessionFeedbackChannel')}
+            standardLabel
+            data={{ cy: 'set-gamification' }}
```
rschlaefli:
```suggestion
            data={{ cy: 'set-feedback-enabled' }}
```

rschlaefli (approved):

PR #3886, opened 2023-09-29
feat: automatically award live session achievements for first, second and third rank

sjschlapbach:


sjschlapbach (commented):
@rschlaefli This PR should be basically ready, the only things left to do are:
- seed the three achievements in the production DB and determine how many points / XP should be awarded for each achievement
- include the IDs of the achievements in the production DB in the code

rschlaefli (changes requested):
See TODOs in code

sjschlapbach:
PR has been replaced with PR #3940

PR #3898, opened 2023-10-05
enhance(apps/frontend-manage): introduce updated print view for feedback channel

sjschlapbach:
This pull request introduces the following fixes and enhancements:
- Feedback-channel print view on evaluation works again
- Feedback-channel print view on running session works again
- Both print views utilize the same component and therefore share a unified layout
- Enhanced use of CSS made it possible to avoid that feedbacks break across pages - this should not happen anymore
- Approach always ensures that the current filtered content will be shown on the print view


BShaq (approved):
LGTM :)

PR #3899, opened 2023-10-08
enhance(apps/backend-docker): select JWT based on origin to allow manage and pwa logins within same browser

rschlaefli:


sjschlapbach (approved):
LGTM :)

PR #3904, opened 2023-10-09
enhance: extend data model for elements structure

rschlaefli:


apps/frontend-manage/src/lib/utils/filters.ts
```diff
@@ -180,7 +179,7 @@ export function sortQuestions(questions: any[], sort: any): any[] {
   return questions
 }
 
-export function processItems(items, filters, sort, index): Question[] {
+export function processItems(items, filters, sort, index): Element[] {
```
sjschlapbach, line 183: As the import for Question from GraphQL is removed and Element is not imported, TS seems to use some general "Element" type, which would contradict our objective of typing these elements more consistent?

sjschlapbach: Additionally, it seems like we don't even have this type even available as an import from GraphQL... Therefore, I'll just highlight all positions in the code where this might become an issue with "ElementImportIssue" so that we find it again to replace it

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -281,10 +280,10 @@ function Index() {
   const [sortBy, setSortBy] = useState('')
 
   const [selectedQuestions, setSelectedQuestions] = useState<
-    Record<number, Question | undefined>
+    Record<number, Element | undefined>
```
sjschlapbach, line 284: ElementImportIssue

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -281,10 +280,10 @@ function Index() {
...
   >({})
 
-  const selectedQuestionData: Record<number, Question> = useMemo(
+  const selectedQuestionData: Record<number, Element> = useMemo(
```
sjschlapbach, line 287: ElementImportIssue

apps/frontend-manage/src/pages/index.tsx
```diff
@@ -582,7 +581,7 @@ function Index() {
                 <QuestionList
                   questions={processedQuestions}
                   selectedQuestions={selectedQuestionData}
-                  setSelectedQuestions={(id: number, data: Question) => {
+                  setSelectedQuestions={(id: number, data: Element) => {
```
sjschlapbach, line 585: ElementImportIssue

apps/frontend-pwa/src/types/app.ts
```diff
@@ -1,4 +1,4 @@
-export enum QuestionType {
+export enum ElementType {
```
sjschlapbach: This type is not consistent anymore with the GraphQL implementation. This might be risky to define types with identical names here, which could also be imported from GraphQL?

packages/graphql/src/public/schema.graphql
```diff
@@ -122,6 +122,40 @@ A date-time string at UTC, such as 2007-12-03T10:15:30Z, compliant with the `dat
...
+  KPRIM
+  MC
+  NUMERICAL
+  QUESTION_FREE_TEXT
+  QUESTION_KPRIM
+  QUESTION_MC
+  QUESTION_NUMERICAL
+  QUESTION_SC
+  SC
+}
```
sjschlapbach, line 160: Does it really make sense to keep both SC and QUESTION_SC as separate types here? Wouldn't it make sense to decide to use one or the other and migrate everything accordingly in order not to introduce some combination, which will then have to be resolved manually later on?

packages/prisma/src/prisma/migrations/20231009210245_question_to_element/migration.sql
```diff
@@ -0,0 +1,33 @@
...
+
+-- AlterEnum
+ALTER TYPE "QuestionType" RENAME TO "ElementType";
+ALTER TYPE "ElementType" ADD VALUE 'CONTENT';
+ALTER TYPE "ElementType" ADD VALUE 'FLASHCARD';
+ALTER TYPE "ElementType" ADD VALUE 'QUESTION_SC';
+ALTER TYPE "ElementType" ADD VALUE 'QUESTION_MC';
+ALTER TYPE "ElementType" ADD VALUE 'QUESTION_KPRIM';
+ALTER TYPE "ElementType" ADD VALUE 'QUESTION_FREE_TEXT';
+ALTER TYPE "ElementType" ADD VALUE 'QUESTION_NUMERICAL';
```
sjschlapbach: Probably we would also have to remove the old keys here, in case we decide to migrate completely to the new convention

packages/prisma/src/prisma/migrations/20231010100715_element_additions_and_practice_quiz/migration.sql
```diff
@@ -0,0 +1,149 @@
+-- CreateEnum
+CREATE TYPE "PublicationStatus" AS ENUM ('DRAFT', 'PUBLISHED');
+
+-- CreateEnum
+CREATE TYPE "ElementOrderType" AS ENUM ('SEQUENTIAL', 'SHUFFLED', 'LAST_RESPONSE');
+
+-- CreateEnum
+CREATE TYPE "ElementInstanceType" AS ENUM ('PRACTICE_QUIZ');
```
sjschlapbach, line 8: Should we also insert live quiz, micro learning and group activity here? They are also already defined as elementStackTypes

packages/prisma/src/prisma/migrations/20231010100715_element_additions_and_practice_quiz/migration.sql
```diff
@@ -0,0 +1,149 @@
...
+
+-- CreateEnum
+CREATE TYPE "ElementStackType" AS ENUM ('LIVE_QUIZ', 'PRACTICE_QUIZ', 'MICROLEARNING', 'GROUP_ACTIVITY');
+
+-- AlterTable
+ALTER TABLE "Element" ADD COLUMN     "version" INTEGER NOT NULL DEFAULT 1;
+
+-- AlterTable
+ALTER TABLE "QuestionResponse" ADD COLUMN     "elementInstanceId" INTEGER,
+ALTER COLUMN "questionInstanceId" DROP NOT NULL;
```
sjschlapbach, line 18: If we just create a new column and drop the old one don't we lose all the data then?

packages/prisma/src/prisma/migrations/20231010100715_element_additions_and_practice_quiz/migration.sql
```diff
@@ -0,0 +1,149 @@
...
+
+-- AlterTable
+ALTER TABLE "QuestionResponseDetail" ADD COLUMN     "elementInstanceId" INTEGER,
+ALTER COLUMN "questionInstanceId" DROP NOT NULL;
```
sjschlapbach, line 22: Same as above

packages/prisma/src/prisma/migrations/20231010100715_element_additions_and_practice_quiz/migration.sql
```diff
@@ -0,0 +1,149 @@
...
+
+-- CreateTable
+CREATE TABLE "ElementInstance" (
+    "id" SERIAL NOT NULL,
+    "originalId" TEXT,
+    "type" "ElementInstanceType" NOT NULL,
+    "elementType" "ElementType" NOT NULL,
+    "order" INTEGER NOT NULL,
+    "options" JSONB NOT NULL,
```
sjschlapbach, line 31:
```suggestion
    "activityOptions" JSONB NOT NULL,
```

packages/graphql/src/types/app.ts
```diff
@@ -66,65 +87,67 @@ export type QuestionOptionsNumerical = {
...
-  type: QuestionType
-  displayMode: QuestionDisplayMode
   content: string
   pointsMultiplier: number
-  explanation: string | null
+  explanation?: string | null
+
+  options: object
+
+  // TODO: these legacy props have been moved to options
```
sjschlapbach, line 113: So we would have to change the `options: object` above to `options: ElementOptions` and save the corresponding values at the correct place during the creation of instances?

packages/graphql/src/types/app.ts
```diff
@@ -66,65 +87,67 @@ export type QuestionOptionsNumerical = {
...
-> extends Question {
+interface IElementData<Type extends ElementType, Options extends ElementOptions>
+  extends Element {
   type: Type
   options: Options
 }
 
-export type ChoicesQuestionData = IQuestionData<
+export type ChoicesElementData = IElementData<
   'SC' | 'MC' | 'KPRIM',
```
sjschlapbach, line 128: Here we might get the first inconsistencies, if we also allow QUESTION_SC, etc. now as DB types. Probably it would make sense to type all of this consistently and migrate from SC to QUESTION_SC from the beginning?

packages/graphql/src/types/app.ts
```diff
@@ -66,65 +87,67 @@ export type QuestionOptionsNumerical = {
...
-  QuestionOptionsChoices
+  ElementOptionsChoices
 >
-export type FreeTextQuestionData = IQuestionData<
+export type FreeTextElementData = IElementData<
   'FREE_TEXT',
```
sjschlapbach, line 132: Same as above

packages/graphql/src/types/app.ts
```diff
@@ -66,65 +87,67 @@ export type QuestionOptionsNumerical = {
...
-  QuestionOptionsFreeText
+  ElementOptionsFreeText
 >
-export type NumericalQuestionData = IQuestionData<
+export type NumericalElementData = IElementData<
   'NUMERICAL',
```
sjschlapbach, line 136: Same as above

packages/graphql/src/types/app.ts
sjschlapbach: Also, do we have to use string values here or could we also use ElementType.Question_FREE_TEXT or something?

packages/graphql/src/types/app.ts
```diff
@@ -152,10 +175,49 @@ export type AllQuestionInstanceTypeData =
...
+  | MicrolearningStackOptions
+  | PracticeQuizStackOptions
+
+interface IElementStackOptions<
+  Type extends ElementStackType,
+  Options extends ElementStackOptions | null
+> {}
+
+export type AllElementStackOptions =
+  | IElementStackOptions<'LIVE_QUIZ', LiveQuizStackOptions>
```
sjschlapbach, line 213: Do we have to use Strings as types here and cannot use more general types like ElementStackType.LIVE_QUIZ?

packages/prisma/src/prisma/schema.prisma
```diff
@@ -120,22 +126,27 @@ model User {
...
-  questions         Question[]
+  questions         Element[]
   mediaFiles        MediaFile[]
   tags              Tag[]
   questionInstances QuestionInstance[]
   sessions          LiveSession[]
   learningElements  LearningElement[]
   microSessions     MicroSession[]
   groupActivities   GroupActivity[]
+  elementInstances  ElementInstance[]
```
sjschlapbach, line 137: Why do we need a link from the user to all ElementInstances?

packages/prisma/src/prisma/schema.prisma
```diff
@@ -153,44 +164,95 @@ model MediaFile {
...
   NUMERICAL
+
+  // new types
+  CONTENT
+  FLASHCARD
+  QUESTION_SC
+  QUESTION_MC
+  QUESTION_KPRIM
+  QUESTION_FREE_TEXT
+  QUESTION_NUMERICAL
```
sjschlapbach: Same as earlier - why do need to keep the old SC / MC / ... types, if we introduce QUESTION_SC, etc. - wouldn't it make sense to directly migrate and remove the old types within the same step?

packages/prisma/src/prisma/schema.prisma
```diff
@@ -199,25 +261,117 @@ model Question {
   updatedAt DateTime @updatedAt
 }
 
-model Tag {
-  id   Int    @id @default(autoincrement())
-  name String
+// TODO: elements are only supported for the new PRACTICE_QUIZ type in the first iteration. add the other types here when working on them.
+enum ElementInstanceType {
+  PRACTICE_QUIZ
```
sjschlapbach, line 268: Could we also already add LIVE_QUIZ, MICROLEARNING and GROUP_ACTIVITY as types here so that we need one migration step less later on? :)

packages/prisma/src/prisma/schema.prisma
```diff
@@ -199,25 +261,117 @@ model Question {
...
-  owner   User   @relation(fields: [ownerId], references: [id], onDelete: Cascade, onUpdate: Cascade)
-  ownerId String @db.Uuid
+  // TODO: moved to options Json
+  // resetTimeDays    Int?
+  // pointsMultiplier Int                      @default(1)
+  // TODO: moved to results Json
+  // participants    Int                      @default(0)
+
+  /// [PrismaElementInstanceOptions]
+  options Json // contains element type specific settings (resetTimeDays for LE instances, pointsMultiplier where relevant)
```
sjschlapbach, line 287:
```suggestion
  activityOptions Json // contains element type specific settings (resetTimeDays for LE instances, pointsMultiplier where relevant)
```

packages/prisma/src/prisma/schema.prisma
```diff
@@ -813,20 +945,147 @@ model Participation {
...
+  participant   Participant @relation(fields: [participantId], references: [id], onDelete: Cascade, onUpdate: Cascade)
+  participantId String      @db.Uuid
+
+  participation Participation?
+
+  sessionParticipation   Participation? @relation("SessionLeaderboards", fields: [sessionParticipationId], references: [id])
+  sessionParticipationId Int?
+
+  sessionBlock   SessionBlock? @relation(fields: [sessionBlockId], references: [id], onDelete: Cascade, onUpdate: Cascade)
+  sessionBlockId Int?
```
sjschlapbach, line 987: Do we link a leaderboard entry to a session block at any point? For the session leaderboard we use the link to the session and for the course leaderboard we use the link to the course. In the production DB, there are zero results with non NULL sessionBlockId

packages/prisma/src/prisma/schema.prisma
```diff
@@ -813,20 +945,147 @@ model Participation {
...
+  bookmarkedElementStacks ElementStack[]
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
 
   @@unique([courseId, participantId])
 }
 
+enum LeaderboardType {
+  SESSION_BLOCK
```
sjschlapbach, line 966: Possibly we could remove the SESSION_BLOCK type from here (see below)

apps/frontend-pwa/src/types/app.ts
rschlaefli: wondering why we even have this here. I do not think that it is used anywhere :D

rschlaefli: no longer used with small adjustments and will remove

packages/graphql/src/types/app.ts
rschlaefli: nope. that will cause a typescript error, as using a type as a value is not possible. so we will have to work hardcoded here

packages/graphql/src/types/app.ts
rschlaefli: no, in base element data (which is used only to extend from), we do not care what is in the options. that will be specified in the type specific interfaces that extend from this here :) so options will stay an object.

packages/graphql/src/types/app.ts
sjschlapbach: We cannot use Prisma types here, that makes sense - what keeps us from using the types from ../ops? There we export an enum with the types as values? However, these would then have to be compatible with Prisma, so it's probably not the best idea and easier to have them hard-coded...

packages/graphql/src/types/app.ts
rschlaefli: in terms of db maintenance, we will need to move these props into the options object, that is correct, but that does not change anything here except that the three properties will disappear from this type and be added to the specific options type.

packages/graphql/src/types/app.ts
rschlaefli: same as previously. types cannot be used here, as TS expects values. in some way we tell it "if the runtime value is LIVE_QUIZ, the corresponding type will be LiveQuizStackOptions". this will allow TS to narrow down to the type when we have an if statement checking for the type

packages/prisma/src/prisma/migrations/20231009210245_question_to_element/migration.sql
rschlaefli: removing these for now

packages/prisma/src/prisma/migrations/20231010100715_element_additions_and_practice_quiz/migration.sql
rschlaefli: for now, I would insert the activities that support elements, and add them later once they are compatible

packages/prisma/src/prisma/schema.prisma
rschlaefli: activityOptions probably does not make sense either, as e.g., the points multiplier could be moved here and that is the combination of the activity and element multiplier. will leave it at options for now.. maybe "derivedOptions" or something could be used..

packages/graphql/src/types/app.ts
sjschlapbach: makes sense, thanks for the explanation :)

packages/prisma/src/prisma/schema.prisma
rschlaefli: same as we have one from user to question instance, maybe because of bookmarks or responses and responsedetails?

rschlaefli: this is added automatically as we have owner/ownerId on the ElementInstance. just the other side of the relation

packages/prisma/src/prisma/schema.prisma
rschlaefli: same as above, I would add it once the activity supports elements

packages/prisma/src/prisma/schema.prisma
rschlaefli: I have no idea, I guess we do not persist anything during a block, it was conceptualized as a storage for LB snapshots after each block, but I guess that does not exist in any code :D we would need to go through code and check before we remove it here though

packages/prisma/src/prisma/schema.prisma
sjschlapbach: For what do we need the owner on the ElementInstance then? We didn't need one on the QuestionInstance?

PR #3912, opened 2023-10-14
chore: add seed script for flashcards

sjschlapbach:
This pull request includes the following relevant changes:
- Modified approach to test question seeding, which now relies on `originalId` instead of the database id
- Basic seeding for flashcards from an XML-input to a practice quiz with instances of the corresponding elements
- Connections between the created instances and the course, which allow to create nicer repetition structure


sjschlapbach:
Remaining potential TODO: pick values, which are relevant for flashcards before creating corresponding instances. This might be especially important here as flashcards only need minimal information beside the content (and nothing like hasAnswerFeedbacks, etc.).

PR #3913, opened 2023-10-20
enhance: add backend components to save student responses to flashcard elements

sjschlapbach:


packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -13,6 +15,16 @@ export async function getPracticeQuizData(
       stacks: {
         include: {
           elements: {
+            include:
+              ctx.user?.sub && ctx.user?.role === UserRole.PARTICIPANT
```
rschlaefli:
```suggestion
              ctx.user?.sub && ctx.user.role === UserRole.PARTICIPANT
```

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -13,6 +15,16 @@ export async function getPracticeQuizData(
...
+                ? {
+                    responses: {
+                      where: {
+                        participantId: ctx.user?.sub,
```
rschlaefli:
```suggestion
                        participantId: ctx.user.sub,
```

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
-  // }
-
-  // if (element.orderType === OrderType.SHUFFLED) {
-  //   return {
-  //     ...element,
-  //     ...instancesWithoutSolution,
-  //     instances: shuffle(instancesWithoutSolution.instances),
-  //   }
-  // }
+  if (ctx.user?.sub && ctx.user?.role === UserRole.PARTICIPANT) {
```
rschlaefli:
```suggestion
  if (ctx.user?.sub && ctx.user.role === UserRole.PARTICIPANT) {
```

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+  const existingInstance = await ctx.prisma.elementInstance.findUnique({
+    where: {
+      id,
+    },
+  })
+
+  if (!existingInstance) return null
+
+  // 0 = wrong, 1 = partial, 2 = correct
+  if (![0, 1, 2].includes(correctness)) {
```
rschlaefli: use an enum here to make it clear what each number says (like in the comment)

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+    return null
+  }
+
+  // TODO: also handle the case where the participant is logged in, but has no valid participation in the considered course
```
rschlaefli: maybe we can ignore participations and related topics completely, at least as long as the flashcards have no impact on leaderboards (just give XP)

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+  if (!ctx.user?.sub) {
+    // TODO: if user is not logged in, do not check the questionResponse, only save it on the instance / update the instance results
```
rschlaefli: I would probably not save anything for anonymous users on flashcards and just continue to the next FC on the frontend without an API request (reduce load)

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+  }
+
+  // update the aggregated data on the element instance
+  await ctx.prisma.elementInstance.update({
+    where: {
+      id,
+    },
+    data: {
+      results: {
+        wrong: existingInstance.results.wrong + (correctness === 0 ? 1 : 0),
```
rschlaefli: use the new enum here and everywhere else

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+      },
+    },
+  })
+
+  // TODO: think about possibility to fuse these two cases with a single upsert (however, keep in mind that we need to update JSON stuff here)
+  if (existingResponse) {
+    let aggregatedResponse =
+      existingResponse.aggregatedResponses as AggregatedResponseFlashcard
+    switch (correctness) {
+      case 0:
```
rschlaefli: if we use an enum and not ints but words for the keys, we could do it generically here and increment using 

...aggregatedResponse,
[response]: aggregatedResponse[response] + 1,
total: aggregatedResponse[response] + 1

and just check precedingly whether the response is one of the three in the enum

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+      },
+      data: {
+        trialsCount: {
+          increment: 1,
+        },
+        response: {
+          correctness,
+        },
+        aggregatedResponses: aggregatedResponse,
+        correctCount: {
```
rschlaefli: would we need partial correct count as well, or is having only correct count sufficient?

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+          increment: correctness === 2 ? 1 : 0,
+        },
+        correctCountStreak: {
+          increment:
+            correctness === 2 ? 1 : -existingResponse.correctCountStreak,
+        },
+        lastCorrectAt: correctness === 2 ? new Date() : undefined,
```
rschlaefli: should probably also update a lastAnsweredAt timestamp, to have both

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+      wrong: 0,
+      partial: 0,
+      correct: 0,
+      total: 0,
+    }
+
+    switch (correctness) {
+      case 0:
+        aggregatedResponse = {
+          wrong: 1,
```
rschlaefli: simplify by just changing the relevant numbers to 1 instead of redefining the entire object

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -27,32 +39,247 @@ export async function getPracticeQuizData(
...
+    const questionResponse = await ctx.prisma.questionResponse.create({
+      data: {
+        trialsCount: 1,
+        response: {
+          correctness,
+        },
+        aggregatedResponses: aggregatedResponse,
+        correctCount: correctness === 2 ? 1 : 0,
+        correctCountStreak: correctness === 2 ? 1 : 0,
+        lastCorrectAt: correctness === 2 ? new Date() : undefined,
```
rschlaefli: set last answered at if not done so by prisma automatically

packages/graphql/src/types/app.ts
```diff
@@ -38,7 +38,23 @@ export type QuestionResponseValue = {
   value: string
 }
 
-export type QuestionResponse = QuestionResponseChoices | QuestionResponseValue
+export type QuestionResponseFlashcard = {
+  correctness: number
+}
+
+export type AggregatedResponse = AggregatedResponseFlashcard
```
rschlaefli: what could be aggregated responses for the other types? we need to think about this when starting with learning analytics

packages/prisma/src/prisma/schema.prisma
```diff
@@ -1011,9 +1011,16 @@ model QuestionResponse {
   lastAwardedAt      DateTime?
   lastXpAwardedAt    DateTime?
```
rschlaefli, line 1014: add lastAnsweredAt DateTime?

rschlaefli (changes requested):

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: @rschlaefli In principle, I totally agree, but if I saw it correctly, the link from the question response table to the "Participation" table is required... Therefore, we either have to move away from this concept (which would probably require rigorous testing of all other use cases) or still require a participation in my opinion.

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: wip implementation has been committed

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: wip implementation has been committed

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: wip implementation has been committed

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: wip implementation has been committed

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: will probably depend heavily on the type of repetition algorithm we use, but to have full flexibility, it would probably make sense to track this number as well

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: 1. Doesn't this kind of coincide with the `updatedAt` timestamp? Even though this one could obviously be kept at the same value, even when changing something else on the same instance
2. If we introduce such a timestamp, the `lastCorrectAt` timestamp might be superfluous or could at least be derived from the `lastAnsweredAt` timestamp and the correct streak and other values?

On the other hand, this would be quick win and we might be able to find an interesting use case for this later on during analysis

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: successfully implemented

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: successfully implemented

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: successfully implemented

rschlaefli (approved):

PR #3914, opened 2023-10-20
refactor(packages/graphql): improve typing and consistency in pothos schema and services

sjschlapbach:
This pull request should resolve typing issues, which were ignored with an `as any` typecast until now.


sjschlapbach:
Even though, the approach of defining custom references and implementations for the types did resolve many of the issues, new ones appeared as soon as the object is returned in a list or certain other issues might also contribute to typing issues.

packages/graphql/src/schema/query.ts
```diff
@@ -256,7 +257,7 @@ export const Query = builder.queryType({
         },
         resolve(_, args, ctx) {
           // FIXME by fixing type issues in LearningElementService
```
sjschlapbach: @rschlaefli Locally, this line does not show any type issues anymore - is it possible that they have been resolved automatically through other changes?

packages/graphql/src/schema/query.ts
```diff
@@ -268,7 +269,7 @@ export const Query = builder.queryType({
...
```
sjschlapbach: As above, this does not show any type issues locally

sjschlapbach (commented):
LGTM

PR #3915, opened 2023-10-21
feat: allow public access to evaluation page using HMAC verified link

rschlaefli:


sjschlapbach (approved):
Added the same links also to the frontend-control embedding page, the implementation works well and looks good to me :)
Eventually, we could also try to fuse together the evaluation modals to reduce code duplication

PR #3939, opened 2023-11-09
enhance(packages/graphql): ensure that group activity clue distribution is more even

sjschlapbach:


packages/graphql/src/services/groups.ts
```diff
@@ -472,14 +472,18 @@ export async function startGroupActivity(
...
-
-      const clueAssignments = group.participants.flatMap((participant, ix) => {
-        const start = ix * cluesPerMember
-        const end = start + cluesPerMember
-        const clues = shuffledClues.slice(start, end)
+      let remainingClues = groupActivity.clues.length
+      let remainingMembers = groupMemberCount
+      let startIx = 0
+
+      const clueAssignments = group.participants.flatMap((participant) => {
```
rschlaefli: refactor to use reduce instead of reassigning variables

rschlaefli (changes requested):

sjschlapbach:
@rschlaefli reduce has been fixed, typing is only partially working with any[] type for the clue accumulator

rschlaefli (approved):

PR #3940, opened 2023-11-09
feat: add automatic live session achievement awarding

sjschlapbach:


sjschlapbach:
@rschlaefli Before we can merge this PR, we need to seed the achievements in the production database and replace the corresponding IDs in the constant definitions at the top of the changes files. The test seed implementation for the achievements should be reusable as it already uses available images, etc.

packages/prisma/src/data/seedTEST.ts
```diff
@@ -506,6 +506,62 @@ async function seedTest(prisma: Prisma.PrismaClient) {
       update: {},
     })
   })
+
+  const goldMedalAchievement = await prisma.achievement.upsert({
```
rschlaefli: run seedAchievements.ts in prisma:setup instead of duplicating the achievements code?

packages/prisma/src/data/seedTEST.ts
```diff
@@ -506,6 +506,62 @@ async function seedTest(prisma: Prisma.PrismaClient) {
...
+    where: { id: 11 },
+    create: {
+      id: 11,
+      name: 'Champion',
+      description: 'Du hast einen ersten Platz in einer Live-Session erreicht.',
+      icon: '/achievements/Champ.svg',
+      type: 'PARTICIPANT',
+      rewardedPoints: 200,
```
rschlaefli: add points/xp to the seedAchievements.ts script

packages/prisma/src/data/seedTEST.ts
sjschlapbach: done

packages/prisma/src/data/seedTEST.ts
sjschlapbach: done

PR #3941, opened 2023-11-09
feat: allow to change questionData in unpublished elements

sjschlapbach:
This pull request adds the possibility to update a question on change in all scheduled elements (scheduled / prepared live sessions, draft practice quizzes and draft microlearnings). Note that this will never be done automatically, but only on request of the user.


sjschlapbach:
Once this is merged, this should also be reflected in comments on the following topics:
https://klicker-uzh.feedbear.com/boards/feature-requests/posts/questions-update-changed-questions-in-planned-session-and-activities
https://community.klicker.uzh.ch/t/update-of-question-in-sessions/117

packages/graphql/src/services/questions.ts
```diff
@@ -446,3 +450,170 @@ export async function getFileUploadSas(
...
+                  learningElement: {
+                    where: {
+                      status: DB.LearningElementStatus.DRAFT,
+                    },
+                  },
+                },
+              },
+            },
+          },
+          microSession: {
```
rschlaefli, line 495: add elementStack for the new approach

sjschlapbach: will be added later as this data is not visible in the question pool at the moment and adaptions of these element do not make sense for a user

rschlaefli (approved):

PR #3967, opened 2023-11-16
fix(packages/prisma): add additional binary target for prisma client

sjschlapbach:


sjschlapbach:
Issue seems to have resolved itself on the v3.1-sprint-nov branch...

PR #3971, opened 2023-11-23
fix: trim usernames and email addresses in all forms and on submission

sjschlapbach:


apps/frontend-pwa/src/components/forms/UpdateAccountInfoForm.tsx
```diff
@@ -87,9 +87,9 @@ function UpdateAccountInfoForm({
         const result = await updateParticipantProfile({
           variables: {
             isProfilePublic: values.isProfilePublic,
-            password: values.password,
-            username: values.username,
-            email: values.email,
+            password: values.password.trim(),
+            username: values.username?.trim() || '',
+            email: values.email?.trim() || '',
```
rschlaefli: setting this to '' instead of null/undefined would not set it to '' in the backend as well? (instead of undefined meaning do not change it)

rschlaefli (approved):

apps/frontend-pwa/src/components/forms/UpdateAccountInfoForm.tsx
sjschlapbach: has been fixed with corresponding adaptations to the mutation typing to be consistent accross frontend and backend

PR #3992, opened 2023-12-14
enhance: hide delegated access creation when using delegated logins

sjschlapbach:


apps/frontend-manage/src/pages/user/settings.tsx
```diff
@@ -181,11 +184,16 @@ function Settings() {
...
-        <Setting title={t('auth.delegatedAccess')}>
-          <Suspense fallback={<Loader />}>
-            <DelegatedAccessSettings shortname={user?.userProfile?.shortname} />
-          </Suspense>
-        </Setting>
+        {typeof scope?.userScope !== 'undefined' &&
+          scope.userScope === UserLoginScope.AccountOwner && (
+            <Setting title={t('auth.delegatedAccess')}>
+              <Suspense fallback={<Loader />}>
+                <DelegatedAccessSettings
```
rschlaefli: I would move the GetUserScope query into delegated access settings so that it is part of the suspense component as well (otherwise you will have loading for the scope query, once that is loaded, the delegated access component will start loading).

rschlaefli (changes requested):

apps/frontend-manage/src/pages/user/settings.tsx
sjschlapbach: Sounds good. Should we still keep it as a separate query and just access it through this function in the GraphQL query or incorporate the scope directly as a return value of the delegated access query?

sjschlapbach: The corresponding query has been moved into the component as requested so that the suspense will handle the entire data loading. Moreover, the query is now executed together with the user endpoint query for simplicity

rschlaefli (approved):

PR #3996, opened 2023-12-24
enhance(cypress): add testing attributes to all buttons, links and other components across all apps

sjschlapbach:


apps/frontend-manage/src/components/sessions/EmbeddingModal.tsx
```diff
@@ -36,13 +36,16 @@ function LazyHMACLink({
   return (
     <div className="flex flex-row items-center gap-3">
       <Link
-        className="px-2 py-1 bg-slate-100 rounded hover:bg-slate-200"
+        className="px-2 py-1 rounded bg-slate-100 hover:bg-slate-200"
         href={link}
         target="_blank"
       >
-        {link}
+        <span data-cy={`open-embedding-link-session-${sessionId}`}>{link}</span>
```
rschlaefli, line 43: Not sure how it will work when using "click" from Cypress on the child element of the Link/a, not the link itself, as it would work with data on the `span`? 

Might be better to do these with legacyBehavior and passHref and using an `a` tag as the child (https://nextjs.org/docs/pages/api-reference/components/link#legacybehavior).

rschlaefli (approved):

apps/frontend-manage/src/components/sessions/EmbeddingModal.tsx
sjschlapbach: @rschlaefli There are multiple examples in the code (e.g. `load-session-list`), which are already used in cypress tests and do not cause problems. Probably Cypress handles these clicks as locations in the viewport and not as clicks on a specific DOM element?
However, to be on the safe side, I refactored this to follow the logic mentioned by you in PR #3998

PR #4028, opened 2024-02-01
chore(packages/prisma): add initial migration scripts from old to new data structure

rschlaefli:


packages/prisma/src/prisma/migrations/20240206151543_element_instance_migration_id/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+  - Made the column `migrationId` on table `ElementInstance` required. This step will fail if there are existing NULL values in that column.
+
+*/
+-- AlterTable
+ALTER TABLE "ElementInstance" ADD COLUMN     "migrationId" INTEGER;
+
+-- UpdateTable
+UPDATE "ElementInstance"
+SET "migrationId" = id
+WHERE "migrationId" IS NULL;
```
sjschlapbach, line 15: Do we need to store the id in the migrationId, if we plan drop this column asap anyway?

packages/prisma/src/scripts/2024-02-01_migrate_learning_elements.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  return {}
+}
+
+function preparePracticeQuizElementData(
+  learningElement: LearningElement,
+  questionInstance: QuestionInstance
+) {
+  return {
+    ...questionInstance.questionData,
```
sjschlapbach, line 38: The underlying `AllElementTypeData` type is only formulated for questions, with a type for flashcards being commented. Should we extend this as well with this PR for completeness?

packages/prisma/src/scripts/2024-02-01_migrate_learning_elements.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+}
+
+function preparePracticeQuizInstanceResults(
+  learningElement: LearningElement,
+  questionInstance: QuestionInstance
+) {
+  return {
+    ...questionInstance.results,
```
sjschlapbach, line 47: Do we want to keep the results in the current form and migrate them later to a more general setup, where the total number of answers is tracked as well, etc.? This change would probably also require backend changes, so it might sense to postpone it, but track the task to be completed before the v3.1 release

packages/prisma/src/scripts/2024-02-01_migrate_learning_elements.ts
```diff
@@ -0,0 +1,223 @@
...
+    const quiz = await prisma.practiceQuiz.upsert({
+      where: {
+        id: elem.id,
+      },
+      create: {
+        id: elem.id,
+        name: elem.name,
+        displayName: elem.displayName,
+        description: elem.description,
+        status: elem.status,
```
sjschlapbach: Even though these enums currently contain the same values, it might make sense to have a proper logic as the types are different?

sjschlapbach:
```suggestion
        status: elem.status === LearningElementStatus.PUBLISHED ? PublicationStatus.PUBLISHED : PublicationStatus.DRAFT
```

packages/prisma/src/scripts/2024-02-01_migrate_learning_elements.ts
```diff
@@ -0,0 +1,223 @@
...
+          connect: {
+            id: elem.owner.id,
+          },
+        },
+
+        stacks: {
+          connectOrCreate: elem.stacks.map((stack) => ({
+            where: {
+              type_practiceQuizId_order: {
+                type: 'PRACTICE_QUIZ',
```
sjschlapbach: Should we maybe use `ElementStackType.PRACTICE_QUIZ` instead of the actual values, to make sure this script could also be run with changed type namings?

packages/prisma/src/scripts/2024-02-06_migrate_group_activity.ts
```diff
@@ -0,0 +1,150 @@
...
+                  where: {
+                    type_migrationId: {
+                      type: ElementInstanceType.GROUP_ACTIVITY,
+                      migrationId: instance.id,
+                    },
+                  },
+                  create: {
+                    migrationId: instance.id,
+                    type: ElementInstanceType.GROUP_ACTIVITY,
+                    elementType: instance.questionData.type,
```
sjschlapbach: Is this type to be trusted / always set properly? Just to be sure that we do not save anything wrong into the new schema from the beginning...

packages/prisma/src/scripts/2024-02-06_migrate_microlearning.ts
```diff
@@ -0,0 +1,191 @@
...
+    const quiz = await prisma.microLearning.upsert({
+      where: {
+        id: elem.id,
+      },
+      create: {
+        id: elem.id,
+        name: elem.name,
+        displayName: elem.displayName,
+        description: elem.description,
+        status: elem.status,
```
sjschlapbach: Same as above for the practice quiz status

packages/prisma/src/scripts/2024-02-06_migrate_microlearning.ts
```diff
@@ -0,0 +1,191 @@
...
+        owner: {
+          connect: {
+            id: elem.owner.id,
+          },
+        },
+
+        stacks: {
+          connectOrCreate: elem.instances
+            // filter old and broken instances without the order
+            .filter((instance) => instance.order !== null)
```
sjschlapbach: Could we not recover those as probably none of the instances has an order and just assign the order according to the occurrence - as they were probably also shown like this in the frontend?

sjschlapbach (approved):
LGTM - just some minor comments / questions

PR #4049, opened 2024-02-17
enhance: restructure course element overview with tabs

sjschlapbach:
Screenshot 2024-02-17 at 22 24 52


sjschlapbach (commented):
@rschlaefli This branch would be ready. Due to the changes of the course overview, some of the tests, where publishing in this view is required will most likely fail. However, as hopefully all new components should already contain the required data-cy attributes, the fix should be easy and fast

rschlaefli (approved):

PR #4139, opened 2024-07-11
rework to JWT and improve UI

rschlaefli:


apps/frontend-pwa/src/components/forms/LoginForm.tsx
```diff
@@ -161,7 +166,18 @@ export function LoginForm({
                     type={passwordHidden ? 'password' : 'text'}
                   />
 
-                  <div className="flex flex-row justify-end w-full">
+                  <div className="flex flex-row justify-between w-full">
```
rschlaefli, line 169: added a back button in case one wants to go back to pw login

packages/graphql/src/builder.ts
```diff
@@ -46,7 +46,7 @@ const builder = new SchemaBuilder<{
   }
 }>({
   authScopes: async (ctx) => ({
-    authenticated: !!ctx.user?.sub,
+    authenticated: !!ctx.user?.sub && ctx.user.scope !== UserLoginScope.OTP,
```
rschlaefli, line 49: this should prevent usage of a magic link for login (it should only be usable to generate a real login token). has not been verified that it works like that yet

packages/graphql/src/services/accounts.ts
```diff
@@ -104,6 +104,27 @@ export function createParticipantToken(participantId: string) {
   )
 }
 
+async function doParticipantLogin(
```
rschlaefli, line 107: tried to extract this stuff into a function as it is done four times now (better naming ideas? :D)

packages/graphql/src/services/accounts.ts
```diff
@@ -150,69 +168,75 @@ export async function sendMagicLink(
...
-  const magicLinkExpiration = dayjs().add(15, 'minute').toDate()
+  const participantData = participantWithUsername[0]
 
-  // set loginToken and loginTokenExpiresAt
-  await ctx.prisma.participant.update({
-    where: { id: participant.id },
-    data: {
-      magicLinkToken: magicLinkToken,
-      magicLinkExpiresAt: magicLinkExpiration,
+  const magicLinkJWT = JWT.sign(
```
rschlaefli, line 187: using a JWT makes it independent of the db, we don't need to track expiration, and it is based on cryptography with our app secret (signed) -> still need to test that an outdated magic link will show some message ("has expired,try again...")

packages/graphql/src/services/accounts.ts
```diff
@@ -150,69 +168,75 @@ export async function sendMagicLink(
...
+      expiresIn: '30 minutes',
+    }
+  )
 
-  // TODO: send email to user with magic link
-  const magicLink = `${process.env.APP_STUDENT_SUBDOMAIN}/magicLogin?token=${magicLinkToken}&username=${participant.username}`
+  const magicLink = `${process.env.APP_STUDENT_SUBDOMAIN}/magicLogin?token=${magicLinkJWT}`
   console.log(magicLink)
 
+  // // TODO: send email to user with magic link
```
rschlaefli, line 203: forgot to install the @types for nodemailer before the flight, but this will be the only remaining thing now (rest is reviewed and tested)

packages/graphql/src/services/accounts.ts
```diff
@@ -150,69 +168,75 @@ export async function sendMagicLink(
...
 }
 
 export async function loginParticipantMagicLink(
-  { username, token }: { username: string; token: string },
+  { token }: { token: string },
   ctx: Context
 ) {
-  if (!username || !token) return null
+  //
+  const tokenData = JWT.verify(token, process.env.APP_SECRET as string) as {
```
rschlaefli, line 217: need to check if it throws an error when it is expired -> if yes, need to try/catch and return false to show "expired message"

packages/i18n/messages/de.ts
```diff
@@ -82,10 +82,10 @@ export default {
       token: 'Token',
       passwordRepetition: 'Passwort (Wiederholung)',
       signin: 'Anmelden',
-      usernameError: 'Bitte geben Sie Ihren Nutzernamen ein.',
+      usernameError: 'Bitte geben Sie Ihren Nutzernamen oder E-Mail ein.',
       passwordError: 'Bitte geben Sie Ihr Passwort ein.',
       studentLoginError:
-        'Benutzername oder Passwort sind falsch. Falls Sie Ihr Passwort vergessen haben, nutzen Sie bitte die "Magic Link Login" Funktion.',
+        'Nutyername/E-Mail oder Passwort sind falsch. Falls Sie Ihr Passwort vergessen haben, nutzen Sie bitte die "Magic Link Login" Funktion.',
```
rschlaefli:
```suggestion
        'Nutzername/E-Mail oder Passwort sind falsch. Falls Sie Ihr Passwort vergessen haben, nutzen Sie bitte die "Magic Link Login" Funktion.',
```

PR #4140, opened 2024-07-15
add @klicker-uzh/transactional with react-email and implement basic email sending using nodemailer

rschlaefli:
TODO: getting it to work with oauth 2 login for Outlook (would send from noreply-klicker@df.uzh.ch) in production (will require ZI cooperation) or using cleverreach with SMTP (would send from klicker@mailings.uzh.ch, less preferable) -> separate PR TBD
TODO: replacing everything with env variables and adding to deploy scripts -> separate PR TBD


sjschlapbach (approved):
LGTM :)

PR #4141, opened 2024-07-15
chore: remove links between question instances and (detail) responses

sjschlapbach:
This pull request makes the following changes to the database scheme (incl. migration):
- Remove links between `QuestionInstance` (only used in live quiz) and `QuestionResponse` and `QuestionResponseDetails` - all corresponding ids are set to null already
- Make the links between `QuestionResponse` / `QuestionResponseDetail` and `ElementInstance` required, since response objects without this link cannot be associated with a question / flashcard / ... anyway and are therefore not really informative - all of the corresponding ids are defined (not response with `null` as `elementInstanceId`)
- Change the deletion / updating logic such that responses are deleted alongside the corresponding `ElementInstance`, since there might be responses without an assigned question otherwise, which does not make sense, since the results cannot be interpreted without the question.

## Summary by CodeRabbit

- **New Features**
	- Enhanced database schema for improved data integrity by making `elementInstanceId` a required field in the `QuestionResponse` and `QuestionResponseDetail` tables.
  
- **Bug Fixes**
	- Removed the unnecessary `questionInstanceId` field from the `QuestionResponse` and `QuestionResponseDetail` models to streamline relationships.
  
- **Refactor**
	- Updated relationships to establish a mandatory connection with `ElementInstance`, enhancing the overall structure of the data model.


rschlaefli (approved):
Agree with the change and makes the schema cleaner and data consistent. However, to be discussed whether we want to delete element instances and thus trigger cascading delete (or whether we remove cascading delete and actually want to have a transaction fail when an instance is deleted but responses are still around?), or whether we keep element instances around even if the element they correspond to is deleted (or even also for quizzes that were published but are deleted?), thus keeping leaderboards and responses consistent (as the instance should contain all data necessary to know what it was about)? Especially if we want to do timelines, audit logs, and such things, we want to make sure that they cannot change after participants have seen them, and responses should be consistent with such derived data (meaning responses should basically never be deleted).

PR #4142, opened 2024-07-15
chore: remove links between leaderboard entry and session blocks

sjschlapbach:
This pull request removes the unused link between the session block and leaderboard entries, since a use of such a connection is not planned in the near future, and we should therefore probably try and keep the schema clean. At the time of creation, no links existed, to verify, run either of the SQL queries below.

ADDITIONALLY: Currently, the leaderboard entries on session cancellation were obtained from a join no the session blocks, where no valid entries were available. This has been fixed to link to the session itself correctly now - corresponding data in the database might be slightly inconsistent (as they have not been removed on session cancellation), but should not cause any problems in the future.

```sql
SELECT * FROM "LeaderboardEntry" JOIN "SessionBlock" ON "LeaderboardEntry"."sessionBlockId" = "SessionBlock".id
```
(should return zero entries)

```sql
SELECT * FROM "LeaderboardEntry" WHERE "sessionBlockId" IS NOT NULL
```
(should return zero entries)


packages/prisma/src/prisma/schema.prisma
```diff
@@ -959,10 +958,6 @@ model LeaderboardEntry {
   sessionParticipation   Participation? @relation("SessionLeaderboards", fields: [sessionParticipationId], references: [id])
```
rschlaefli, line 957: should we also remove SESSION_BLOCK from the LeaderboardType enum if we are not going to use it?

rschlaefli (approved):

packages/prisma/src/prisma/schema.prisma
sjschlapbach: absolutely, good idea

sjschlapbach: has been fixed

PR #4146, opened 2024-07-16
fix(apps/frontend-manage): disable reinitialization of form as this is not required

sjschlapbach:
This reinitialization setting for Formik caused the session creation / editing wizards to reset at random points when selecting different filters and/or questions. Since we handle the case where loaded data to edit a certain element before even creating the form, the correct initial values should be available from the beginning and no reinitialization should be required at any point. Tests to reproduce the resetting issue indicated its resolution.

### DIFF
--- a/apps/frontend-manage/src/components/sessions/creation/MultistepWizard.tsx
+++ b/apps/frontend-manage/src/components/sessions/creation/MultistepWizard.tsx
@@ -163,7 +163,6 @@ function MultistepWizard({
       onSubmit={handleSubmit}
       validationSchema={step.props.validationSchema}
       isInitialValid={initialValid}
-      enableReinitialize
     >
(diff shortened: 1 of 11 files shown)


apps/frontend-manage/src/components/sessions/creation/MultistepWizard.tsx
```diff
@@ -163,7 +163,6 @@ function MultistepWizard({
       onSubmit={handleSubmit}
       validationSchema={step.props.validationSchema}
       isInitialValid={initialValid}
-      enableReinitialize
```
sjschlapbach, line 166: This is the prop, which caused the random bug, reinitializing the form at the wrong moment. Through a corresponding conditional statement, the cases where data is loaded for editing are handled before the initialization of the form. Therefore, there is no need for reinitialization.

PR #4172, opened 2024-08-15
fix(packages/graphql): ensure that questions in live session are loaded in correct order

sjschlapbach:


rschlaefli (approved):
wow, really was just a stupid bug :D

PR #4178, opened 2024-08-16
implement account activation and improve code structure

rschlaefli:


sjschlapbach (commented):
Awesome, generation of activation links and their use seem to work perfectly :)

PR #4179, opened 2024-08-16
chore: split prisma schema into multiple files

rschlaefli:


sjschlapbach (approved):
This makes it much more readable :)

PR #4180, opened 2024-08-16
enhance: add status for review workflow to elements, improve tag list layout, add sorting by modification date

rschlaefli:
# ElementStatus

added ElementStatus to enable review workflows

image

TODO:

- [x] add filters in the sidebar for the status?
- [x] cover setting the status in cypress

# Prisma Optimize

Also added prisma optimize in dev mode and if the PRISMA_OPTIMIZE flag in doppler is set

this enables tracing of all db interactions

see 
https://www.prisma.io/blog/prisma-optimize-early-access

TODO

- [ ] think about tracing in production (is that useful / do we want it, e.g., with opentelemetry and jaeger) -> e.g., like https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/opentelemetry-tracing


packages/i18n/messages/en.ts
```diff
@@ -835,7 +844,9 @@ Since the KlickerUZH app is not yet available on the iOS App Store, follow these
       EDITTitle: 'Edit question',
       DUPLICATETitle: 'Duplicate question',
       questionType: 'Question type',
+      questionStatus: 'Status',
       selectQuestionType: 'Select question type',
+      selectQuestionStatus: 'Select status',
```
sjschlapbach, line 852: The translations for `questionStatus` and `selectQuestionStatus` seem to be missing in the `de.ts` file (unfortunately, the GH UI does not allow making this suggestion directly since it is not part of the diff...)

packages/prisma/src/prisma/migrations/20240816193204_element_status/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- CreateEnum
+CREATE TYPE "ElementStatus" AS ENUM ('DRAFT', 'IN_REVIEW', 'READY');
```
sjschlapbach: (detail) Should we maybe just call it `REVIEW` instead of `IN_REVIEW` to keep notation simple?

packages/i18n/messages/de.ts
```diff
@@ -6,6 +6,15 @@ export default {
         'Der eingegebene Wert ist keine Zahl oder liegt nicht im vorgegebenen Bereich.',
       ftPlaceholder: 'Ihre Antwort...',
     },
+    DRAFT: {
+      statusLabel: 'Entwurf',
+    },
+    IN_REVIEW: {
+      statusLabel: 'Zum Review',
```
sjschlapbach: (detail) Similar as in the database, we could simply call this `Review`? Otherwise, I would rename it to `Im Review` to be consistent with the English version

packages/i18n/messages/en.ts
```diff
@@ -6,6 +6,15 @@ export default {
         'The entered value is not a number or is not in the specified range.',
       ftPlaceholder: 'Your answer...',
     },
+    DRAFT: {
+      statusLabel: 'Draft',
+    },
+    IN_REVIEW: {
+      statusLabel: 'In Review',
```
sjschlapbach: (detail) Possibly rename as suggested in the database file

packages/prisma/src/prisma/migrations/20240816193204_element_status/migration.sql
```diff
@@ -0,0 +1,5 @@
...
+
+-- AlterTable
+ALTER TABLE "Element" ADD COLUMN     "status" "ElementStatus" NOT NULL DEFAULT 'DRAFT';
```
sjschlapbach: (most critical?) Using the `DRAFT` state as a default here in the migration would probably mean that all existing questions would automatically be assigned to the draft state? This would probably confuse most users 😅

sjschlapbach (changes requested):
Nice to have these new statuses available in the UI and database. Added some comments regarding details / the migration, which can just be resolved if not agreed with.

- [ ] Possibly it would make sense though to extend the drag & drop / selection operations and/or accepted elements by the session creation wizards to only accept elements with the `READY` state? Or should this status only serve as an indication for people sharing an account?
- [ ] Additionally, it would be pretty quick and simple, but still nice, to include the selection of different states in the test suite as well to ensure that these settings work as expected? -> already included in PR description

packages/prisma/src/prisma/migrations/20240816193204_element_status/migration.sql
rschlaefli: As written in the description:

_default in DB is currently set to draft -> when deploying, we should change everything already in the DB to ready. maybe this could also be done right in the migration (create the field with a default of READY and then change the default to DRAFT after within the same migration)_

however, right now I would also tend to just change it to READY always. because in the UI I also made it ready when creating an element for people who do not care. main idea with DRAFT as default was, e.g., for AI generated questions, where we want them as drafts. but we will just set draft manually there :)

sjschlapbach: Totally missed all the description below the image in the PR description, sorry. Setting it to READY by default and only manually setting questions to the DRAFT state when generating them (or possibly also to REVIEW when importing them?) would make total sense :)

PR #4184, opened 2024-08-19
feat: add possibility for students to provide feedback to questions

sjschlapbach:
This pull request introduces the additional possibility for students to upvote and downvote (as well as comment) any questions in element stacks in asynchronous elements. The corresponding contributions are stored in a separate table and linked for learning analytics analyses.

Screenshot 2024-08-19 at 18 30 49


packages/graphql/src/services/participants.ts
```diff
@@ -486,11 +486,44 @@ export async function flagElement(
...
+    },
+  })
+
   if (
     !elementInstance?.elementStack.practiceQuiz?.course?.notificationEmail &&
     !elementInstance?.elementStack.microLearning?.course?.notificationEmail
   ) {
-    return null
+    // return early if no notification email has been specified -> only set database entry
+    return 'OK'
```
sjschlapbach, line 524: @rschlaefli We should probably return a successful result here, since the feedback has been stored in the database at this point. Before, we simply showed an error toast in the frontend whenever no notification email was specified in the database...

apps/frontend-pwa/src/components/practiceQuiz/InstanceHeader.tsx
```diff
@@ -20,12 +28,43 @@ interface InstanceHeaderProps {
...
   instanceId,
+  elementId,
   name,
   withParticipant,
   correctness,
   className,
 }: InstanceHeaderProps) {
+  const [rateElement] = useMutation(RateElementDocument)
   const [modalOpen, setModalOpen] = useState(false)
+  const [vote, setVote] = useState(0) // TODO: optionally fetch last rating from DB
```
sjschlapbach, line 62: Should we fetch the previous submitted feedback here? The added index should allow to fetch the corresponding entry relatively fast with the participantId and elementInstanceId being known?

apps/frontend-pwa/src/components/practiceQuiz/InstanceHeader.tsx
```diff
@@ -20,12 +28,43 @@
...
+
+    if (res.data?.rateElement?.upvote) {
+      setVote(1)
+    } else if (res.data?.rateElement?.downvote) {
+      setVote(-1)
+    } else {
+      // TODO: show error!
+      setVote(0)
+    }
+  }
```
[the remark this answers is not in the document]
sjschlapbach: Has been addressed with the most recent commit

apps/frontend-pwa/src/components/practiceQuiz/InstanceHeader.tsx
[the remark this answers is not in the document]
sjschlapbach: This is subject to one of the comments already part of this PR

packages/graphql/src/services/participants.ts
rschlaefli: yes 👍

apps/frontend-pwa/src/components/practiceQuiz/InstanceHeader.tsx
rschlaefli: if we do that, we should do it in a scalable way and not send x frontend requests in a stack with x items. instead fetch it once for the entire stack by sending all ids of the elements in the stack. or probably even better already in the request where we get the questions for the stack from DB (we can join and inside the join add a where clause with the two elements of your index, instance id and participant id?) 

-> if we do it like that, it might be relatively efficient. but not a large priority to have it I think, as it is probably not used so often anyways

packages/graphql/src/schema/mutation.ts
```diff
@@ -450,13 +451,27 @@ export const Mutation = builder.mutationType({
...
           return ParticipantService.flagElement(args, ctx)
         },
       }),
 
+      rateElement: t.withAuth(asParticipant).field({
+        nullable: true,
+        type: ElementFeedback,
+        args: {
+          elementInstanceId: t.arg.int({ required: true }),
+          elementId: t.arg.int({ required: true }),
```
rschlaefli, line 467: why do we need both ids to rate it? is the elementId not inside the instance already?

rschlaefli (approved):
looks really cool, very nicely integrated in UI!

apps/frontend-pwa/src/components/practiceQuiz/InstanceHeader.tsx
sjschlapbach: --> will be implemented on separate PR

PR #4186, opened 2024-08-19
enhance: properly refetch element feedback after mutation to update student UI

sjschlapbach:


sjschlapbach:
Potential workaround for this problem could be to fetch the previous element feedbacks only on the current stack (instead of the current practice quiz / microlearning / group activity). In this way, it might be possible to properly update the cache...

PR #4187, opened 2024-08-19
enhance: implement LTI 1.3 for PWA, add session-storage based approach for LTI in iframe, and link existing accounts if email is provided

rschlaefli:


sjschlapbach (approved):
LGTM

PR #4205, opened 2024-08-23
feat: compute instance statistics (correctness, average timeliness, etc.) on stack submission

sjschlapbach:
This pull request introduces the computation of instance statistics in both the case where a participant is signed in and the case where a student participates anonymously. The following fields are filled:
- `correctCount` / `partialCorrectCount` / `wrongCount` for logged in participants on practice quizzes / microlearnings
- `first`- and `last`- versions of the correctness counts only for logged in participants and only on practice quizzes
- `uniqueParticipantCount` resembling the number of participants with login that took part in a voting on this question (should be equal to the number of linked question responses as long as the participant does not delete his account)
- `averageTimeSpent` the mean of all average time values on the question response by logged in participants
- The upvoting and downvoting parameters are not affected by this PR, but handled separately through the corresponding rating function.

Additionally, this PR contains a refactor of the `respondToQuestion` function, which should be more efficient now and require less unnecessary queries / only the necessary joins.


packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -1259,12 +1445,10 @@ export async function respondToQuestion(
     !treatAnonymous &&
     ctx.user.role === UserRole.PARTICIPANT
   ) {
-    const hasPreviousResponse = instance?.responses?.length > 0
-
-    if (hasPreviousResponse) {
+    if (!!existingResponse) {
```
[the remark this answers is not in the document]
sjschlapbach: has been addressed in new commit

packages/graphql/src/services/practiceQuizzes.ts
```diff
@@ -247,6 +251,126 @@ function updateSpacedRepetition({
   }
 }
 
+function computeNewAverageTimes({
+  existingInstance,
+  existingResponse,
+  answerTime,
+}: {
+  existingInstance: ElementInstance & {
+    instanceStatistics: InstanceStatistics | null
```
rschlaefli, line 260: probably would be better to say the param can't be null and solving the type issue outside than | null and then ! on it?

rschlaefli (approved):

packages/graphql/src/services/practiceQuizzes.ts
sjschlapbach: this seemed pretty much impossible to do due to the join we performed here... but I can try at a later point, made a note

PR #4224, opened 2024-08-30
enhance(apps/frontend-manage): display LTI embed links on course overview

rschlaefli:
Introduce functionality for copying LTI access links to clipboard. Small restyling of all element tiles on course overview (move icon before header, move tags to bottom right and keep primary action and dropdown on top right to have consistent alignment). Tried to make the link logic reusable.

For now, the link is displayed based on catalyst. This should be changed to only supported orgs based on affiliation once that is available. The documentation will state that the platform needs to be set-up with us before LTI can be used.

image

image


sjschlapbach (approved):
LGTM - we should make sure that we track the affiliation of the user in the future to properly make UZH-only features (like LTI links) only available to these

PR #4276, opened 2024-09-24
fix: create a participation when logging in via LTI from a page within the course directory, redirect to new microlearning page with courseId param

rschlaefli:


sjschlapbach (approved):
LGTM

PR #4290, opened 2024-10-04
feat(apps/frontend-manage): add search option for tags list

TheHummel:
## Summary by CodeRabbit

- **New Features**
	- Introduced a search input feature in the Suspended Tags component, allowing users to filter tags based on their input.
	- Enhanced tag rendering to display only those matching the search criteria.

- **Chores**
	- Introduced a new reverse proxy service for improved routing in the Docker setup.
	- Updated network configuration for the reverse proxy service to enhance connectivity.
	- Corrected indentation for consistency in the Docker configuration.

- **Dependencies**
	- Updated the `remark-math` dependency version in the documentation project from `~5.1.1` to `~6.0.0`.


apps/frontend-manage/src/components/questions/tags/SuspendedTags.tsx
```diff
@@ -46,55 +72,70 @@ function SuspendedTags({ showUntagged, activeTags, handleTagClick }: Props) {
...
+    <div>
+      <TextField
+        placeholder={t('manage.general.searchPlaceholder')}
+        value={searchInput}
+        onChange={(newValue: string) => {
+          setSearchInput(newValue)
+          handleSearch(newValue)
+        }}
+        icon={faMagnifyingGlass}
+        className={{
```
sjschlapbach, line 84: Could we maybe update the styling of the search field a bit so that it does not take as much space as it does right now? (we only have very limited space here...) I would go for small text (as the tags have this text size as well), reduce the margins a bit and center the search icon in the field.
Also after running the code, I'm relatively confident, that the w-30 attribute can be removed, since it does not seem to have an effect :)

apps/frontend-manage/src/components/questions/tags/SuspendedTags.tsx
```diff
@@ -46,55 +72,70 @@ function SuspendedTags({ showUntagged, activeTags, handleTagClick }: Props) {
...
+              handleTagClick={(tag: string) =>
+                handleTagClick({
+                  tagName: tag,
+                  isTypeTag: false,
+                  isStatusTag: false,
+                  isUntagged: false,
+                })
+              }
+              active={activeTags.includes(tag.name)}
+              onMoveDown={
```
sjschlapbach, line 104: Maybe we should disable and hide the sorting functionality here whenever the search is active, since sorting does not have any visible effect in these cases?

sjschlapbach (changes requested):
Works great! Also very cool that we do not have any problems with stopwords in this search :)
After a few minor styling changes, this should be ready to be merged

apps/frontend-manage/src/components/questions/tags/SuspendedTags.tsx
```diff
@@ -27,11 +31,33 @@ interface Props {
...
+  const filteredTags = useMemo(() => {
+    if (data?.userTags) {
+      const search = new JsSearch.Search('id')
+      search.searchIndex = new JsSearch.TfIdfSearchIndex('id')
+      search.indexStrategy = new JsSearch.AllSubstringsIndexStrategy()
+      search.addIndex('name')
+      search.addDocuments(data.userTags)
+
+      if (searchInput) {
+        return search.search(searchInput) as Tag[]
```
sjschlapbach, line 52: Typecasting should only be required here - the two `as Tag[]`s below can be removed to make sure that the standard tag type is still correct

sjschlapbach: unfortunately, proper typing for this library does not seem to be available otherwise...

apps/frontend-manage/src/components/questions/tags/SuspendedTags.tsx
TheHummel: not quite sure what you mean. for me the manual "move up" and "move down" buttons work when search is active

sjschlapbach (approved):
LGTM :)

PR #4292, opened 2024-10-04
fix(apps/frontend-pwa): ensure that LTI workflows are only triggered inside LMS environment

sjschlapbach:


apps/frontend-pwa/src/lib/getParticipantToken.ts
```diff
@@ -23,13 +23,18 @@ export default async function getParticipantToken({
...
-  //     participantToken,
-  //     cookiesAvailable: !!cookies['participant_token'],
-  //   }
-  // }
+  // TODO: only check for existing participantToken once participation issues with LTI are resolved
+  if (
+    participantToken &&
+    !cookies['lti-token'] &&
+    !query.jwt &&
+    !(req.method === 'POST')
```
sjschlapbach, line 31: @rschlaefli Do you think we can already switch back to only checking whether a participationToken exists at this point? The current logic should avoid that the early return is triggered if any of the LTI workflows would be followed afterward and works for students that are normally logged in through the pwa application.

rschlaefli: yes, we should be able to switch back now

PR #4320, opened 2024-10-19
refactor: use union types for element data, question data and evaluation results

sjschlapbach:


packages/graphql/src/graphql/ops/FElementData.graphql
```diff
@@ -1,14 +1,13 @@
...
   elementData {
-    id
-    elementId
-    name
-    type
-    content
-    explanation
-    pointsMultiplier
-
     ... on ChoicesElementData {
```
sjschlapbach, line 3: @rschlaefli  Do you know if there is a way to extract these shared attributes somewhere to only define them once? Unfortunately, the change to use union types for proper frontend types does not seem to support shared fields on these union types in the Pothos Schema and leaving them at the top level (as before) results in build errors for the graphql package

PR #4333, opened 2024-10-24
enhance: add possibility to check for exact solution on numerical questions

sjschlapbach:
This branch enables the validation for exact solutions in the following way:
- Validation that min and max on solution range are not identical is disabled
- Grading logic has been updated to include the float precision range around a solution as correct
- Evaluation view has been updated to mark a solution as "exact correct solution" if min and max are less than 2 float precisions apart

## Release Notes

- **New Features**
  - Enhanced validation logic for numerical questions, simplifying the handling of minimum values.
  - Improved rendering of solution ranges with clearer checks for displaying correct solutions.

- **Bug Fixes**
  - Adjusted floating-point precision checks in grading logic to ensure accurate response evaluations.

- **Localization Updates**
  - Added and refined translation strings for both German and English, improving user guidance and error messages across the application.

### DIFF
--- a/apps/frontend-manage/src/components/sessions/evaluation/QuestionEvaluation.tsx
+++ b/apps/frontend-manage/src/components/sessions/evaluation/QuestionEvaluation.tsx
@@ -190,7 +190,22 @@ function QuestionEvaluation({
                 </UserNotification>
               )}
               {showSolution &&
-                currentInstance.questionData.options.solutionRanges && (
+                currentInstance.questionData.options.solutionRanges &&
+                (currentInstance.questionData.options.solutionRanges.length ===
+                  1 &&
+                currentInstance.questionData.options.solutionRanges[0].min &&
+                currentInstance.questionData.options.solutionRanges[0].max &&
+                currentInstance.questionData.options.solutionRanges[0].min >
+                  currentInstance.questionData.options.solutionRanges[0].max -
+                    2 * Number.EPSILON ? (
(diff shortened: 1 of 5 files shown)


apps/frontend-manage/src/components/sessions/evaluation/QuestionEvaluation.tsx
```diff
@@ -190,20 +190,35 @@ function QuestionEvaluation({
                 </UserNotification>
               )}
               {showSolution &&
-                currentInstance.questionData.options.solutionRanges && (
+                currentInstance.questionData.options.solutionRanges &&
+                (currentInstance.questionData.options.solutionRanges.length ===
```
rschlaefli, line 194: what if we have multiple correct solutions? would it be simple to do it as a mapping here? (check if the constraints apply to all elements in the array and then just display them all as a list, assuming for now that mixing the approaches is not possible)

rschlaefli (changes requested):

apps/frontend-manage/src/components/sessions/evaluation/QuestionEvaluation.tsx
sjschlapbach: will be done separately once types are available (together with extended error parsing for solution ranges). Otherwise this branch would require extensive testing that everything works and with types all these extensions will be a lot easier (this is the fallback for deployment)

PR #4338, opened 2024-10-28
refactor: migrate live quiz to unified data structure (i.e., from question instances/blocks to element block and element instances)

sjschlapbach:
DO NOT SQUASH THIS BRANCH ON MERGE - development branch for live quiz migration

Before / after merging and applying this branch, execute the following steps:
- [ ] Test that all functionalities of live quiz (including sessions with time limit, element instance updates, etc.) work correctly
- [ ] Verify that variable point logic (timeToZeroBonus and maxBonusPoints) still works as it should
- [ ] Migration script should: copy group activity status to new status, copy element stack type to new type, migrate QuestionInstance to ElementInstance, migrate SessionBlock to ElementBlock, migrate LiveSession to LiveQuiz
- [ ] After migration: Remove QuestionInstance, SessionBlock, LiveSession, rename new group activity status to old one and drop old enum, rename new element stack type to old one and drop old column, (drop originalId columns?!)
- [ ] Create database dump to avoid any potential data loss
- [ ] Apply all database migrations


apps/auth/package.json
```diff
@@ -17,10 +17,8 @@
     "next": "15.0.0",
     "next-auth": "4.24.8",
     "next-intl": "3.21.1",
-    "nookies": "2.5.2",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "sharp": "0.33.5",
```
rschlaefli, line 23: I believe this is a prod dependency especially in standalone deployments but this might have changed with Next.js 15. I'll add it back and we can check it later :)

apps/frontend-manage/package.json
```diff
@@ -83,6 +71,7 @@
     "eslint-config-next": "~15.0.0",
     "postcss": "~8.4.47",
     "postcss-import": "~16.1.0",
+    "react-tagcloud": "~2.3.3",
```
rschlaefli: ! Why did we move this production dependency to dev deps? Would fail once deployed, as in Docker, we only get "dependencies" installed.

apps/frontend-manage/package.json
```diff
@@ -22,40 +22,30 @@
...
     "dayjs": "1.11.13",
     "deepmerge": "4.3.1",
     "formik": "2.4.6",
     "generate-password": "1.7.1",
     "graphql": "16.9.0",
     "graphql-codegen-persisted-query-ids": "0.2.0",
     "graphql-ws": "5.16.0",
-    "html-to-image": "1.11.11",
     "is-hotkey": "0.2.0",
-    "js-cookie": "3.0.5",
```
rschlaefli, line 35: ! I think this is necessary as we are using it in shared-components (useStickyState) and the component is imported in frontend-manage.

apps/frontend-manage/package.json
```diff
@@ -22,40 +22,30 @@
...
     "js-search": "2.0.1",
-    "localforage": "1.10.0",
```
rschlaefli: ! Same, used in useStickyState and therefore used in frontend-manage

apps/frontend-manage/package.json
```diff
@@ -22,40 +22,30 @@
...
     "lodash": "4.17.21",
     "nanoid": "5.0.8",
     "next": "15.0.0",
     "next-intl": "3.21.1",
-    "nookies": "2.5.2",
     "react": "18.3.1",
-    "react-csv-downloader": "3.1.1",
```
rschlaefli, line 44: ! this is also used in shared-components (DataTable).

rschlaefli (approved):
Will merge this now as functionality has been verified and/or fixed. Will be tested further on staging including the other PRs before deployment tomorrow.

Small TODOs for before/after deployment that I saw during testing:
- Fix realtime feedback charts on quiz cockpit not displayed nicely
- Fix countdown block closing but not updating lecturer UI (still showing with 0 sec)
- Fix grid display layout of choice questions

PR #4362, opened 2024-11-06
enhance(apps/frontend-manage): make activity evaluation pages resizable

TheHummel:
## Summary by CodeRabbit

- **New Features**
	- Introduced resizable and collapsible panels in the Choices, FT, and NR Evaluation components, enhancing user interactivity.
	- Implemented local state management for sidebar collapse functionality, improving user control over panel visibility.

- **Bug Fixes**
	- Improved layout responsiveness and user experience by replacing static structures with dynamic resizing capabilities.


sjschlapbach (approved):
Looks great! However, I would wait with merging this branch until we have completed the architectural changes of the live quiz (PR #4338), where we combine the live quiz and asynchronous activity evaluation. There, we should also make sure that the QR codes are properly shown on the sidebars and do not cause scrolling or similar undesired behavior.

apps/frontend-manage/src/components/evaluation/elements/ChoicesEvaluation.tsx
```diff
@@ -18,29 +25,43 @@ function ChoicesEvaluation({
   chartType,
   showSolution,
 }: ChoicesEvaluationProps) {
+  const [isCollapsed, setIsCollapsed] = useLocalStorage(
```
rschlaefli: https://www.npmjs.com/package/react-resizable-panels

would be better to use the autosaveId prop as shown here instead of doing it with a manual hook :) this also persists the size used for the resizeable areas instead of just the collapsed state

apps/frontend-manage/src/components/evaluation/elements/ChoicesEvaluation.tsx
```diff
@@ -18,29 +25,43 @@ function ChoicesEvaluation({
...
-        )}
+      </ResizablePanel>
+      <ResizableHandle withHandle />
+      <ResizablePanel
+        key={`parent-${instanceEvaluation.id}-${isCollapsed ? 'collapsed' : 'expanded'}`}
+        defaultSize={isCollapsed ? 0 : 20}
+        minSize={10}
+        collapsible
+        collapsedSize={0}
+        onCollapse={() => setIsCollapsed(true)}
```
rschlaefli: this should no longer be needed as it handles collapsing and size persistence automatically

apps/frontend-manage/src/components/evaluation/elements/FTEvaluation.tsx
```diff
@@ -18,27 +25,56 @@ function FTEvaluation({
   chartType,
   showSolution,
 }: FTEvaluationProps) {
+  const [isCollapsed, setIsCollapsed] = useLocalStorage(
```
rschlaefli: same

apps/frontend-manage/src/components/evaluation/elements/NREvaluation.tsx
```diff
@@ -18,29 +25,45 @@ function NREvaluation({
   chartType,
   showSolution,
 }: NREvaluationProps) {
+  const [isCollapsed, setIsCollapsed] = useLocalStorage(
```
rschlaefli: same

rschlaefli (changes requested):
very cool, small change to use the feature directly implemented in resizeable instead of the custom hook

apps/frontend-manage/src/components/evaluation/elements/FTEvaluation.tsx
TheHummel: used useState instead now to be able to hide sidebar because else part of content is still visible in collapsed state

rschlaefli (approved):
Very nice! Made a small adjustment so the state is actually saved per question type and not for a specific instance id (that will probably not be opened a second time anyway). Also to not fill up local storage with so many key/value pairs.

LGTM and will merge after the related live quiz branch is integrated.

PR #4372, opened 2024-11-12
[DELETE - DEBUGGING ONLY]: add scripts to identify instances with null choices or missing ix in choices options

sjschlapbach:


sjschlapbach:
Corresponding script fixing these issues has been merged and applied to the production database in the meantime

PR #4384, opened 2024-12-03
[PoC] chore: try to aggregate weekly and monthly analytics from daily analytics

sjschlapbach:
The aggregation does not seem to work with the currently tracked information, since the responseCount cannot be updated correctly (it is now known, which questions have been answered on which day) and computing the mean counts for longer time horizons from their daily values would result in potentially significant distortions of the data


sjschlapbach:
Exiting computation based on question response details is kept

PR #4576, opened 2025-03-21
feat: make base points on elements optional and update corresponding grading logic

sjschlapbach:
After merging this branch, we need to update all existing element instances to contain the relevant information. First apply the database migration:
```bash
cd packages/prisma
pnpm run prisma:deploy:prod
```

Then run the following script:
```bash
cd packages/graphql
pnpm run script ./src/scripts/2025-03-21_add_base_points_instances.ts
```

To verify the success of the script on the production database, run the following command:
```sql
SELECT * FROM "ElementInstance" WHERE "options"->>'basePoints' IS NULL;
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated scoring section in question forms for toggling base points and point multipliers.
  - Updated scoring settings with improved tooltips and localized labels for clearer guidance.

- **UI Enhancements**
  - Streamlined question content display with clearer notifications when points are not awarded.
  - Simplified interface by removing redundant solution settings for a better user experience.

- **Improvements**
  - Enhanced consistency in point calculation across quizzes and live sessions for more reliable scoring.


packages/graphql/src/scripts/2025-03-21_add_base_points_instance_options.ts
```diff
@@ -0,0 +1,42 @@
...
+  for (let i = 0; i < batches; i++) {
+    console.log(`Processing 100-instance batch ${i + 1}/${batches}`)
+
+    const instances = await prisma.elementInstance.findMany({
+      skip: i * 100,
+      take: 100,
+    })
+
+    for (const instance of instances) {
+      await prisma.elementInstance.update({
```
rschlaefli: I would just add a check that basePoints are only set to true if the property does not exist yet in the options (with any value). There should be no case matching that ever, but just to make sure, also if we ran this again in the future for any reason :)

rschlaefli (approved):

PR #4695, opened 2025-05-17
chore: support local https using traefik

rschlaefli:
this adds instructions and traefik configuration to use local SSL/TLS certificats so we can use both http:// and https:// in local development. we might have to decide on one or change other things if we want both possibilities, as there are redirect URLs etc. that are using http:// right now. will also have to verify whether https:// also works on win/linux before we make this default.

this would allow us to develop with HTTP/2, testing more modern things like subscriptions via SSE. mainly built this for jakob who had some issues with subscriptions over SSE in GBL :)

instructions added to https://github.com/uzh-bf/klicker-uzh/wiki/Local-Development#https-setup-with-mkcert-and-traefik-for-http2-support

to evaluate, follow the instructions to generate local certs, restart browsers and app dependencies, and check whether apps are available on both http:// and https:// (they might not work yet because redirect url is http:// etc., but they should be available and not result in gateway errors)

## Summary by CodeRabbit

- **New Features**
  - Enabled HTTPS support for all services routed through the reverse proxy, with explicit TLS certificate configuration and HTTP/2 support.
  - Added access to the Traefik dashboard for easier reverse proxy monitoring.
  - Updated development environments and launch configurations to use HTTPS URLs for enhanced security.

- **Chores**
  - Updated configuration files to standardize reverse proxy settings and SSL certificate mounting.
  - Improved `.gitignore` to exclude the `ssl/` directory.


sjschlapbach (approved):
The setup seems to work well (tested with Mac and Chrome), only added the small step to docs that the browser keychain password might have to be entered during setup

PR #4713, opened 2025-05-23
feat(apps/frontend-manage): add possibility to open answer collection edit modal directly from element edit modal

sjschlapbach:
## Summary by CodeRabbit

- **New Features**
  - Enabled inline editing capabilities for answer collections within forms and selection components.
  - Added an inline edit button for answer collections, conditionally active based on permissions.
  - Enhanced modals and forms to support inline editing and automatic data refresh post-editing.

- **Improvements**
  - Improved handling of answer collection permissions, providing clear messages for insufficient rights.
  - Refined data fetching and state management for answer collections, decoupling from internal queries.
  - Introduced logic to mark shared and editor status for answer collections, preventing duplication.

- **Bug Fixes**
  - Removed logic that reset solutions when changing or removing case study items, simplifying reset behavior.

- **Localization**
  - Added new translation strings for collection selection prompts and permission error messages in English and German.


sjschlapbach:
The implementation approach probably needs to be changed here such that the answer collection edit modal is rendered outside the element edit form context to avoid nested formik forms --> submission does not work. A corresponding trigger function could then be passed down to the element edit form component to control the open state of the answer collection edit form modal

PR #4716, opened 2025-05-26
fix(packages/graphql): make sure that emails for failed final group assignments are only sent once per course

sjschlapbach:
Once this solution has been confirmed to be sufficient for the discussed problematic of sending too many emails (daily) in case random groups could not be formed, we should also add the corresponding email to the production database

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where course owners could receive repeated email notifications about failed automatic group assignments when only one participant is present.


sjschlapbach:
--> potentially failing group activity workflow in CI (before last branch update) was verified to pass locally in final version before merging

PR #4747, opened 2025-06-23
feat(apps/olat-api): define API for OLAT user activity queries

jabbadizzleCode:
## Summary by CodeRabbit

* **New Features**
  * Introduced the OLAT API service, providing REST endpoints for OLAT LMS integration, including course and activity retrieval, activity type listings, and user authentication via API key.
  * Added comprehensive OpenAPI documentation, static activity type definitions, and robust error handling with standardized responses.
  * Implemented rate limiting and health check endpoints.
  * Provided Docker, Kubernetes, and Helm configurations for deployment in production and QA environments.
  * Added server-side validation and redirection for live quiz access in the frontend.
  * Exposed a GraphQL query to validate availability of live quizzes.

* **Documentation**
  * Added detailed user and developer documentation for the OLAT API, including endpoint descriptions, integration flow, deployment, and security guidelines.
  * Updated repository overview documentation to include the OLAT API component.

* **Tests**
  * Implemented automated tests covering all major API endpoints, error cases, and rate limiting.
  * Added test data, helpers, and Docker-based test orchestration.

* **Chores**
  * Integrated CI/CD workflows for building, testing, and deploying the OLAT API service.
  * Updated configuration files to support new deployment and environment variables.
  * Added new TypeScript configurations, build tooling, and package metadata for the OLAT API.


apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+import { pick } from 'remeda'
+import { fileURLToPath } from 'url'
+import { validate as validateUUID } from 'uuid'
+
+const app: express.Express = express()
+const prisma = new PrismaClient()
+const PORT = process.env.PORT || 3000
+
+const API_NAME = 'olat-api'
+const API_KEY = process.env.OLAT_API_KEY
```
rschlaefli, line 28: Agree with the AI here. Verify that it is not empty and fail

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+import path from 'path'
+import { pick } from 'remeda'
+import { fileURLToPath } from 'url'
+import { validate as validateUUID } from 'uuid'
+
+const app: express.Express = express()
+const prisma = new PrismaClient()
+const PORT = process.env.PORT || 3000
+
+const API_NAME = 'olat-api'
```
rschlaefli: what is an API_NAME needed for?

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+const API_KEY = process.env.OLAT_API_KEY
+
+enum StatusCode {
```
rschlaefli: extract the types to a separate file (e.g., types)

apps/olat-api/test/helpers.ts
```diff
@@ -0,0 +1,394 @@
+import {
```
rschlaefli, line 1: @sjschlapbach such a file is probably useful whenever we do tests with docker and dbs? could we share it with other packages somehow?

apps/olat-api/test/index.test.ts
```diff
@@ -0,0 +1,645 @@
...
+  await testInitialization(prisma)
+})
+
+afterAll(async () => {
+  await testCleanup(prisma)
+  await prisma.$disconnect()
+})
+
+function getExpectedResponse(
+  nLQ: number,
```
rschlaefli: understandable naming preferred

bruno/https---localhost-3020-hello.bru
```diff
@@ -0,0 +1,11 @@
+meta {
```
rschlaefli: remove, as no longer valid

.github/workflows/test-olat-api.yml
```diff
@@ -0,0 +1,39 @@
+name: Test OLAT-API functionalities
+
+on:
+  push:
+    branches: ['v3', 'v3*']
+    paths:
+      - 'packages/graphql/**'
```
rschlaefli: why do we have to run tests if graphql changes?

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+}
+
+interface ActivityTypeOfCourse {
+  id: string
+  title: string
+  olatConfigurationKey: string
+  isSubselectionRequired: boolean
+}
+
+let activityTypesAvailable: ActivityType[] = []
```
rschlaefli: we should not have mutable variables at the global level, if needed to do this, try to find another solution that is immutable

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+  'live-quiz',
+  'practice-quiz',
+  'micro-learning',
+] as const // NOTE: add more if required
+let activityKeysGeneral: string[] = []
+type ActivityTypeSubselection = (typeof activityKeysSubselection)[number]
+
+async function apiKeyMiddleware(req: Request, res: Response, next: () => void) {
+  const apiKey = req.headers['x-api-key']
+  if (!apiKey || typeof apiKey !== 'string') {
```
rschlaefli, line 36: it would be better if the server fails instead of it running "healthily" but failing every customer request if it is our fault (not providing an api key at all :))

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+app.use(express.json())
+
+app.get('/health', (req: Request, res: Response) => {
+  res.status(StatusCode.SUCCESS).json({
+    status: 'OK',
+    timestamp: new Date().toISOString(),
+  })
+})
+
+app.get('/', (req: Request, res: Response) => {
```
rschlaefli: this is not needed, as we have the healthcheck endpoint

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+  res.json({ message: 'OLAT API server' })
+})
+
+// Serve OpenAPI specification
+app.get('/openapi.yaml', async (req: Request, res: Response) => {
+  try {
```
rschlaefli, line 81: very basic ai implementation of file serving for the openapi yaml file. we could replace this with the static file serving plugin for express, or a better version that caches the file contents on startup.

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+  '/api-docs',
+  apiReference({
+    url: '/openapi.yaml',
+    theme: 'default',
+  })
+)
+
+function checkHeader(req: Request): StatusCode {
+  const contentType = req.headers['content-type']
+  if (!contentType || typeof contentType !== 'string') {
```
rschlaefli: not sure we actually care about the content type so much that we should fail the request so strictly. I would just default to returning JSON, we don't support anything else.

so I would fail with bad request if anything other than application/json is defined, but otherwise always return json (also if content type is undefined)

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+    return StatusCode.BAD_REQUEST
+  }
+  if (contentType !== 'application/json') {
+    return StatusCode.UNSUPPORTED_MEDIA_TYPE
+  }
+
+  return StatusCode.SUCCESS
+}
+
+async function getCourses(
```
rschlaefli: from a design perspective I would find it strongly preferable to have the service code in a separate module rather than interleaved with the API surface. so I would move these functions out into services.ts and import and use them here, leaving no prisma interactions in the index.ts.

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+        providerAccountId: providerAccountId,
+      },
+    },
+    include: {
+      user: {
+        include: {
+          courses: true,
+        },
+      },
+    },
```
rschlaefli: it might be good, as we join in the user here, to actually also use the select operators from prisma. just so there will never be a chance that we expose anything beside the properties we want to (even if we change code further down).

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+  const coursesDetails = courses.map((course) => ({
+    id: course.id,
+    title: course.name,
+  }))
+
+  return coursesDetails
+}
+
+app.get('/api/configuration/courses', (req: Request, res: Response) => {
+  const headerStatusCode = checkHeader(req)
```
rschlaefli: this should probably be moved to a middleware instead of being repeated in all endpoints

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+  }
+  const coursesDetails = courses.map((course) => ({
+    id: course.id,
+    title: course.name,
+  }))
+
+  return coursesDetails
+}
+
+app.get('/api/configuration/courses', (req: Request, res: Response) => {
```
rschlaefli: it is crucial that we move any endpoint that receives parameters to a POST request. this is important for security and follows best practice. so getting the list of activity types is clearly still a GET request, but submitting a user id and getting the users courses needs to be POST, otherwise the user id could be exposed.

see e.g. https://stackoverflow.com/questions/26671599/are-security-concerns-sending-a-password-using-a-get-request-over-https-valid

rschlaefli: using POST would probably require the addition of body-parser to parse JSON contents of post bodies

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+        return res
+          .status(StatusCode.NOT_FOUND)
+          .json({ error: 'No courses found for this user' })
+      }
+
+      res.set('Content-Type', 'application/json')
+      return res.status(StatusCode.SUCCESS).json({
+        courses: courses,
+        timestamp: new Date().toISOString(),
+        api: API_NAME,
```
rschlaefli: not sure why this would be helpful

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+      console.error('Error fetching activity types:', error)
+      return res
+        .status(StatusCode.INTERNAL_SERVER_ERROR)
+        .json({ error: 'Internal server error' })
+    })
+})
+
+async function getCourseActivityTypes(courseId: string): Promise<any[] | null> {
+  const course = await prisma.course.findUnique({
+    where: { id: courseId },
```
rschlaefli: we also need to make sure that this user can access the course with the specified ID. so we need to ensure that there are derived permissions for the user to access the course. this means also passing the provider account id here via a POST request.

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+
+  const mapSubselection: Record<string, any[]> = {
+    // NOTE: modify if required
+    'live-quiz': liveQuizzes,
+    'practice-quiz': practiceQuizzes,
+    'micro-learning': microLearnings,
+  }
+  const activityKeysGamification = ['course-leaderboard'] // NOTE: modify if required
+
+  const activityTypes: ActivityTypeOfCourse[] = activityTypesAvailable.reduce(
```
rschlaefli: quickly ask AI whether this could be refactored in a more understandable way

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+      })
+  }
+)
+
+async function getActivities(
+  courseID: string,
+  activityTypeKey: ActivityTypeSubselection // NOTE: modify if required
+): Promise<any[] | null> {
+  const course = await prisma.course.findUnique({
+    where: { id: courseID },
```
rschlaefli: same as before, need to also include and verify the user

apps/olat-api/src/index.ts
```diff
@@ -0,0 +1,488 @@
...
+    }
+    const activityTypeKey = req.params.activityTypeKey
+    if (!activityTypeKey || typeof activityTypeKey !== 'string') {
+      return res
+        .status(StatusCode.BAD_REQUEST)
+        .json({ error: 'Missing activityTypeKey' }) // should not happen
+    }
+    if (
+      activityKeysSubselection.indexOf(activityTypeKey) !== -1 &&
+      validateUUID(courseID)
```
rschlaefli: it has been validated before, no?

rschlaefli (changes requested):

apps/olat-api/src/index.ts
jabbadizzleCode: Shouldn't it then be UNSUPPORTED_MEDIA_TYPE instead of BAD_REQUEST?

PR #4755, opened 2025-06-25
enhance: migrate cronjobs and scheduled tasks to new hatchet-based setup, improve scripting and cypress setup, migrate to vitest, dependency upgrades

sjschlapbach:
DEVELOPMENT BRANCH - DO NOT SQUASH ON MERGE

## Summary by CodeRabbit

- **New Features**
  - Schedule and unpublish Live Quizzes from Manage (scheduling modal, UI labels, validation).

- **Improvements**
  - Centralized background scheduling/handlers for timed publish/end of quizzes, microlearnings, and activities.
  - Reverse proxy now exposes host port 8090; improved cross-platform startup/cleanup flow and more robust Redis-backed pub/sub and caching.

- **Tests**
  - GraphQL tests migrated to Vitest; Cypress split/parallel runs and a local test runner added.

- **Documentation**
  - Local testing guide and GraphQL README added.

- **Chores**
  - Tooling upgrades (Node/pnpm), dependency bumps, Docker build optimizations, CI workflow concurrency and caching refinements.


apps/hatchet-worker-general/src/index.ts
```diff
@@ -1,13 +1,75 @@
...
+
+import { createRedisEventTarget } from '@graphql-yoga/redis-event-target'
+import { hatchetClient, prepareHatchetTasks } from '@klicker-uzh/hatchet'
+import EventEmitter from 'events'
+import { createPubSub } from 'graphql-yoga'
+import { Redis } from 'ioredis'
 
 async function main() {
-  const workflows = prepareHatchetTasks(hatchet)
+  const redisExec = new Redis({
```
rschlaefli, line 63: initialization should happen at the root (e.g., start of main function) -> as it is the same thing in backend, all workers, etc., we might also extract this into some kind of shared util function at some point..

apps/hatchet-worker-general/src/index.ts
```diff
@@ -1,13 +1,75 @@
...
+    redisExec,
+    redisCache,
+  })
 
-  const worker = await hatchet.worker('hatchet-worker-general', {
-    // TODO: figure out some way to share tasks with backend (e.g., try shared tasks package)
-    // basic structure according to https://github.com/hatchet-dev/hatchet-typescript-quickstart/tree/main/monorepo
-    workflows,
+  const worker = await hatchetClient.worker('hatchet-worker-general', {
+    workflows: [
```
rschlaefli: need to decide whether and how we would split this up into multiple workers. we could also do it "smartly" and pass the names of the workflows that should be loaded from the dict by a worker from an ENV variable. then we can configure the tasks in the deployment config and scale dynamically instead of hardcoding it.

apps/backend-docker/src/index.ts
```diff
@@ -6,6 +6,7 @@ import { withOptimize } from '@prisma/extension-optimize'
 // import '@sentry/tracing'
 import { createInMemoryCache, type Cache } from '@envelop/response-cache'
 import { createRedisCache } from '@envelop/response-cache-redis'
+import { hatchetClient, prepareHatchetTasks } from '@klicker-uzh/hatchet'
```
rschlaefli, line 8: the hatchetClient that is imported here is already initialized with env variables etc. as a singleton pattern (similar to how prisma works when imported) -> we dont need to duplicate the hatchet client init stuff anymore

apps/hatchet-worker-response-processor/src/index.ts
```diff
@@ -34,9 +34,15 @@ export const processAuthenticatedResponseTask = hatchet.durableTask({
...
 
 async function main() {
-  const worker = await hatchet.worker('hatchet-worker-response-processor', {
-    workflows: [processAuthenticatedResponseTask, processAnonymousResponseTask],
-  })
+  const worker = await hatchetClient.worker(
+    'hatchet-worker-response-processor',
+    {
+      workflows: [
+        processAuthenticatedResponseTask,
```
rschlaefli: need to think whether temporary participants should be treated differently or not. not sure right now but these workflows will change a lot anyways

apps/response-api/src/index.ts
```diff
@@ -130,10 +130,24 @@ async function handleAddResponse(req: IncomingMessage, res: ServerResponse) {
     )
   }
 
-  const cookie =
-    typeof req.headers['cookie'] === 'string'
-      ? req.headers['cookie']
-      : undefined
+  // Only forward participant-related cookies. If both exist, include both.
+  let cookie: string | undefined
```
rschlaefli, line 106: we need to check that the manager cookie will not be forwarded with the message and that we are not counted as authenticated with the manage cookie but as anonymous. hence this "filtering" of cookies before forwarding. could certainly be done nicer, but the ai way works for now (and this entire thing will be rewritten in Go anyway)

packages/graphql/src/lib/util.ts
```diff
@@ -51,25 +50,6 @@ export function formatDate(dateTime: Date) {
   }
 }
 
-export async function sendTeamsNotifications(scope: string, text: string) {
```
rschlaefli, line 54: moved to notifications as this is what it actually does

packages/graphql/src/schema/practiceQuiz.ts
```diff
@@ -176,6 +176,11 @@ export const ElementStack = ElementStackRef.implement({
 export interface IPracticeQuiz
   extends Pick<
     DB.PracticeQuiz,
+    | 'startedCount'
```
rschlaefli: @sjschlapbach here I was not sure when merging, if this is wrong please correct. there was often a "three way merge" and it was a bit chaotic, but types work that way.

packages/graphql/src/services/accounts.ts
```diff
@@ -254,7 +254,7 @@ export async function sendMagicLink(
       templateName: 'MagicLinkRequested',
       variables: { LINK: magicLink },
     },
-    ctx
+    ctx.prisma
```
rschlaefli, line 256: not sure if now it would actually be better to just import prisma everywhere and remove it from ctx altogether (now that it is a singleton in the prisma package). we can check and maybe do this separately though :)

packages/graphql/src/services/notifications.ts
```diff
@@ -148,58 +205,21 @@ export async function sendPushNotifications(_ctx: Context) {
...
-//       } else {
-//         throw error
-//       }
-//     }
-//   }
-// }
+export async function sendTeamsNotifications(scope: string, text: string) {
+  if (process.env.TEAMS_WEBHOOK_URL) {
+    try {
+      return await axios.post(process.env.TEAMS_WEBHOOK_URL, {
```
rschlaefli, line 223: TODO: migrate to new adaptive card mechanism (review how it was done in the course cms)

packages/graphql/src/index.ts
```diff
@@ -34,6 +34,17 @@ import './schema/subscription.js'
 //   })
 // }
 
+export {
```
rschlaefli: these handlers are exported as they are needed in hatchet tasks. backend graphql or a worker will pass them down into the hatchet package. that way the two are completely decoupled (graphql is not installed in hatchet and also not the other way around).

packages/hatchet/src/index.ts
```diff
@@ -0,0 +1,481 @@
+import type { HatchetClient } from '@hatchet-dev/typescript-sdk'
```
rschlaefli: we should probably split this into modules

util/_create_hatchet_token.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
rschlaefli, line 1: moved to util

util/_create_ssl_certificates.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
rschlaefli, line 1: created new script so we don't have to generate with mkcert anymore. automatically called on macos with the run app deps script

_run_app_dependencies.sh
```diff
@@ -1,7 +1,78 @@
 #!/bin/sh
```
rschlaefli, line 1: consolidated all run scripts into one that detects the platform. to be tested on windows and linux but it looks much more sensible than separate scripts (especially now that we have additional logic that would also be repeated)

package.json
```diff
@@ -47,7 +47,7 @@
     "check:format": "lint-staged",
     "check:lint": "turbo run lint --parallel --continue",
     "check:syncpack": "syncpack lint",
-    "dev": "doppler run --config dev -- pnpm run dev:raw",
+    "dev": "CONFIG=dev ./util/_run_with_doppler.sh pnpm dev:raw",
```
rschlaefli, line 49: will first use normal doppler if it works, but if not will fall back to the locally stored token (e.g., useful for external drives)

apps/backend-docker/src/index.ts
```diff
@@ -1,11 +1,12 @@
 import { createRedisEventTarget } from '@graphql-yoga/redis-event-target'
-import { enhanceContext, schema } from '@klicker-uzh/graphql'
```
rschlaefli, line 2: We export the "handlers" from Graphql -> functions that have to be available to any hatchet task from the graphql package. Reasoning -> we dont want to install graphql in hatchet and also not the other way around, as they are independent services and should be decoupled. The same way we provide hatchet tasks to graphql through the prepareHatchetTasks function and pass them down in context.

apps/hatchet-worker-general/src/index.ts
rschlaefli: implemented one approach to do so using an env variable like

HATCHET_WORKFLOWS=sendPushNotifications,...,...

packages/graphql/src/index.ts
```diff
@@ -34,8 +36,28 @@ import './schema/subscription.js'
...
+} from './services/notifications.js'
+import { updateWeeklyTimelineEntries } from './services/participants.js'
+
 export const schema = builder.toSchema({
   schemaDirectives: {
     // oneOf: upperDirectiveTransformer,
   },
 })
+
+export const handlers: HatchetHandlers = {
```
rschlaefli, line 64: this will make sure that we also update the types whenever we add handlers

rschlaefli:
CleanShot 2025-09-01 at 21 30 15

CleanShot 2025-09-01 at 21 34 08

Everything is now deployed to staging including the two workers. Many random issues with GH Actions/PNPM/Types already solved, but also many still remaining..

Response processing works with Hatchet but I believe the Functions are still broken.

packages/transactional/out/MagicLinkRequested.html
```diff
@@ -10,7 +10,7 @@
     * {
```
rschlaefli, line 10: no idea why it reformatted these now

.tool-versions
```diff
@@ -1,2 +1,2 @@
-node 20.18.1
+node 20.19.4
 pnpm 9.8.0
```
rschlaefli, line 2:
```suggestion
pnpm 10.15.0
```

docker-compose.build.yml
```diff
@@ -0,0 +1,45 @@
+services:
```
rschlaefli, line 1: to evaluate whether all containers build locally togehter with the util/_local_build.sh

packages/graphql/src/schema/practiceQuiz.ts
sjschlapbach: fixed it :)

cypress/cypress/e2e/J-elements-free-text-workflow.cy.ts
```diff
@@ -1,33 +1,24 @@
...
+  afterEach(function () {
+    if (this.currentTest.state === 'failed') {
+      Cypress.stop()
+    }
+  })
+
+  it('CLEANUP', () => {
+    cy.cleanup()
+    cy.seed()
+  })
```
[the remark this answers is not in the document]
rschlaefli: This does not work as Cypress runs the before hook also on origin change thus cleaning the DB after a part of the workflow when e.g. switching to the student app. So this change was on purpose.

PR #4769, opened 2025-07-10
chore: remove database structure for live sessions and question instances

sjschlapbach:
## Migration Script (Cache Updates)

```bash
cd package/graphql && pnpm run script:prod ./src/scripts/2025-07-10_drop_live_session_cache_data.ts
```

## Summary by CodeRabbit

* **Chores**
  * Removed all legacy and migration-related fields, models, and enums associated with live sessions and the previous "Question" concept from the database schema and codebase.
  * Cleaned up demo data seeding, templates, tests, and utility scripts to eliminate migration ID generation and usage.
  * Added a script for listing and deleting obsolete live session cache entries.
  * Improved handling of average time spent statistics to ensure default values are always set.
  * Updated build script to set production environment variable for optimized builds.
  * Enhanced type checking by adding a dedicated workflow step for the API application.


packages/graphql/src/scripts/2025-07-10_drop_live_session_cache_data.ts
```diff
@@ -0,0 +1,24 @@
...
+    tls: process.env.REDIS_TLS ? {} : undefined,
+  })
+
+  // drop all live session redis cache entries starting with 's:'
+  const keys = await redisExec.keys('s:*')
+  console.log(`Deleting ${keys.length} keys:`)
+  console.log(keys)
+
+  // TODO: uncomment this line to execute redis pipeline
+  // await redisExec.del(...keys)
```
rschlaefli:
```suggestion
  // await redisExec.unlink(...keys)
```

I think the unlink command is more efficient as it is not blocking (redis will just do it at some point with garbage cleaning)

rschlaefli (approved):
Verified locally & LGTM

PR #4806, opened 2025-07-28
enhance(apps/frontend-manage): extend activity details with student preview

jabbadizzleCode:
## Summary by CodeRabbit

* **New Features**
  * Introduced a detailed, interactive activity details view with points breakdown tables, stack accordions, and instance selection for activities and quizzes.
  * Added new components for displaying stack details, points tables, outdated element warnings, and student element previews.
  * Enhanced localization with new English and German strings for activity details and points.

* **Improvements**
  * Activity details UI is now more structured, responsive, and informative, providing clear points calculations and sample solution indicators.
  * Updated points calculation logic for all activity types, displaying detailed breakdowns at activity, stack, and instance levels.
  * Cypress tests now verify points calculations and UI interactions for all activity types, with new helper commands for asserting points.
  * Simplified student element preview component by delegating rendering to a new wrapper component managing feedback toggling.
  * Updated test suites to consistently set browser locale to English for reliable test execution.

* **Bug Fixes**
  * Ensured consistent browser locale (English) during Cypress tests for reliable language-dependent assertions.

* **Chores**
  * Added metadata to test fixtures to support new activity details features.
  * Minor style adjustment for feedback display in shared components.
  * Exported constants for points per group activity element and per instance for broader usage.
  * Improved test automation by expanding activity detail accordions before assertions.


packages/prisma/src/data/seedTEST.ts
```diff
@@ -103,7 +103,7 @@ export const PARTICIPANT_GROUP_IDS = [
   '11c06c89-0cb4-4d8e-b052-b711f327b8c4',
 ]
 
-const USER_ID_TEST = '73c018c2-d7ad-4844-84b0-814e33420423'
+const USER_ID_TEST = '76047345-3801-4628-ae7b-adbebcfe8821'
```
sjschlapbach: Is this change necessary? Otherwise, we should try not to modify this ID, since it would require all developers with a current setup to re-seed their database

packages/i18n/messages/de.ts
```diff
@@ -1018,6 +1018,17 @@ Da die KlickerUZH-App noch nicht im iOS-App-Store verfügbar ist, folgen Sie die
       catalystRequired:
         'Catalyst-Zugriff erforderlich. Mehr Informationen unter <link></link>.',
       elementPreview: 'Elementvorschau: {element}',
+      elementPreviewRedirect: 'Vorschau in einem neuen Tab öffnen',
+      pointsOverviewDescription: 'Punkteübersicht',
+      elementTypDescription: 'Typ',
```
sjschlapbach:
```suggestion
```

packages/i18n/messages/en.ts
```diff
@@ -1017,6 +1017,17 @@ Since the KlickerUZH app is not yet available on the iOS App Store, follow these
       catalystRequired:
         'Requires catalyst access. For more information, see <link></link>.',
       elementPreview: 'Element Preview: {element}',
+      elementPreviewRedirect: 'Open preview in a new tab',
+      pointsOverviewDescription: 'Points Overview',
+      elementTypDescription: 'Type',
```
sjschlapbach:
```suggestion
```

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -2,22 +2,43 @@ import { useQuery } from '@apollo/client'
...
+  ShadcnTableHeader,
+  ShadcnTableRow,
+  UserNotification,
+} from '@uzh-bf/design-system'
 import { useTranslations } from 'next-intl'
 import Link from 'next/link'
-import { useMemo } from 'react'
+import { useMemo, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
+import StudentElementPreviewActivityDetails from '~/components/questions/manipulation/StudentElementPreviewActivityDetails'
```
sjschlapbach: for consistency with conventions across project :)

```suggestion
import StudentElementPreviewActivityDetails from '../../questions/manipulation/StudentElementPreviewActivityDetails'
```

packages/graphql/src/services/activities.ts
```diff
@@ -425,18 +634,74 @@ export async function getGroupActivityDetails(
...
+        correctnessPoints: null,
+        bonusPoints: null,
+        totalPoints: totalPoints,
+        instance: instance,
+      }
+    }),
   }))
 
-  return { id: groupActivity.id, stacks }
+  const metadata = {
```
sjschlapbach: There is a bug in the computation of available points for a group activity:
Screenshot 2025-07-28 at 16 31 32

apps/frontend-manage/src/components/questions/manipulation/StudentElementPreviewActivityDetails.tsx
```diff
@@ -0,0 +1,190 @@
...
+    })
+
+  // hook running on every instance change to initialize the student response correctly
+  useSingleStudentResponse({
+    instance: instance,
+    setStudentResponse,
+  })
+
+  // memoized stack storage for evaluation components
+  const stackStorage = useMemo<StackStudentResponseType | undefined>(() => {
```
sjschlapbach: Is it correct that the definition of the stackStorage here and the variable `explanationOrFeedbacksDefined` are obtained in the exact same way as in the `StudentElementPreview`? If this is the case, we should try to combine the corresponding logic to avoid code duplication making maintainability even harder

apps/frontend-manage/src/components/questions/manipulation/StudentElementPreviewActivityDetails.tsx
```diff
@@ -0,0 +1,190 @@
...
+  ])
+
+  if (!instance) {
+    return <Loader />
+  }
+
+  return (
+    <div className="max-w-full" data-cy="student-element-preview">
+      <div className="rounded border p-4">
+        {explanationOrFeedbacksDefined && (
```
sjschlapbach: As above, we should also try to extract this to a custom component and share it between this component and the `StudentElementPreview`, unless there are major differences, which I have not noticed?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -57,13 +79,43 @@ function ActivityDetailsModal({
...
+  }, [stacks, selectedInstance])
+
   return (
     <Modal
       open
       loading={loading}
       title={t('manage.activities.activityDetails')}
       onClose={onClose}
-      className={{ content: 'w-96 min-w-96 max-w-96' }}
+      className={{ content: 'w-256 min-w-256 max-w-256' }}
```
sjschlapbach: We should make sure that either the full width is used when an element instance is selected or the width is flexible based on the content - we could try something like `w-max`, if the element preview is assigned a fixed width (as well as the placeholder when no preview is shown)

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+              <span className="font-bold">
+                {t('manage.activityWizard.displayName')}:
+              </span>{' '}
+              {activityDetails?.activityDetails?.metadata.displayName}
+            </li>
+            <li>
+              <h3 className="font-bold">
+                {t('manage.general.pointsOverviewDescription')}:
+              </h3>
+              <ShadcnTable>
```
sjschlapbach: Question: Does it make sense to show the complete table here also for asynchronous activities, where we always only have a single type of points - also not really "base points"? For the data structure it should be alright just to use the `basePoints` for these points, but in the UI it might be confusing

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                          activityDetails?.activityDetails?.metadata
+                            .totalCorrectnessPoints
+                        }{' '}
+                        P.
+                      </ShadcnTableCell>
+                    </ShadcnTableRow>
+                  ) : null}
+                  {activityDetails?.activityDetails?.metadata
+                    .totalBonusPoints &&
+                  activityDetails?.activityDetails?.metadata
```
sjschlapbach: In these cases, it sometimes makes sense to test for `typeof activityDetails?.activityDetails?.metadata.pointsMultiplier !== "undefined"`, since some of the features might break if we decide to introduce a zero multiplier at some point

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                      <ShadcnTableCell className="text-right">
+                        {
+                          activityDetails?.activityDetails?.metadata
+                            .totalCorrectnessPoints
+                        }{' '}
+                        P.
+                      </ShadcnTableCell>
+                    </ShadcnTableRow>
+                  ) : null}
+                  {activityDetails?.activityDetails?.metadata
```
sjschlapbach: Similar to the comment on the pointsMultiplier below, we should probably test if this is `typeof blabla == "undefined"` or `=== null`, since setting the bonus points to zero in the live quiz settings would make this row disappear?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                            .totalBonusPoints
+                        }{' '}
+                        P.
+                      </ShadcnTableCell>
+                    </ShadcnTableRow>
+                  ) : null}
+                </ShadcnTableBody>
+                <ShadcnTableFooter>
+                  <ShadcnTableRow>
+                    <ShadcnTableCell colSpan={1}>
```
sjschlapbach:
```suggestion
                    <ShadcnTableCell colSpan={1} className="font-bold">
```

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
-            </div>
-            <div>
-              {stack.elements.map((instance, instanceIx) => (
-                <Link
-                  href={`/instances/${instance.id}`}
-                  className="text-sm hover:text-slate-700"
-                  key={instance.id}
-                  legacyBehavior
-                  passHref
+      <div className="w-228 h-128 flex flex-row gap-4">
```
sjschlapbach: We should probably try to keep the height flexible - if the user has a high screen, there is no reason not to use this space and avoid scrolling on either the element list or the instance preview on the right side?

sjschlapbach: To make sure this works, we might also need to override some styling on the modal content, not sure about this

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                  >
+                    <FontAwesomeIcon icon={faArrowUpRightFromSquare} />
+                    <span className="italic">
+                      {t('manage.general.elementPreviewRedirect')}
+                    </span>
+                  </div>
+                </a>
+              </Link>
+
+              <Link
```
sjschlapbach: I think the link to the relevant grading section should probably be placed somewhere next to the metadata section of the activity, where the points are actually combined and multipliers shown?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
           )}
         </UserNotification>
```
sjschlapbach: Unfortunately I can't suggest a change here directly, since this component of the file was not changed, but we should add some small margin at the bottom of the user notification if it is shown, it is very close to the text otherwise

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
-                      data-cy={`stack-${index}-instance-${instanceIx}`}
-                    >
-                      <div className="flex flex-row items-center gap-1.5">
-                        {outdatedInstances.includes(instance.id) ? (
-                          <FontAwesomeIcon
-                            icon={faExclamationTriangle}
-                            className="text-uzh-red-100"
-                          />
+                  <div key={stack.id} className="w-full">
+                    <AccordionTrigger className="flex w-full flex-row items-center justify-between p-2">
```
sjschlapbach: The default underline style looks a bit weird here with three distinct texts being underlined - maybe we could override it to disable the underline style at least for the number of participants and the number of points?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                      </ShadcnTableCell>
+                    </ShadcnTableRow>
+                  ) : null}
+                </ShadcnTableBody>
+                <ShadcnTableFooter>
+                  <ShadcnTableRow>
+                    <ShadcnTableCell colSpan={1}>
+                      {t('manage.general.totalPointsDescription')}
+                    </ShadcnTableCell>
+                    <ShadcnTableCell className="text-right text-violet-500">
```
sjschlapbach:
```suggestion
                    <ShadcnTableCell className="text-right text-violet-500 font-bold">
```

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                        </ShadcnTableHead>
+                      </ShadcnTableRow>
+                    </ShadcnTableHeader>
+                    <ShadcnTableBody>
+                      <ShadcnTableRow>
+                        <ShadcnTableCell>
+                          {t('manage.general.basePointsDescription')}{' '}
+                          {activityDetails?.activityDetails?.metadata.type !==
+                          ActivityType.LiveQuiz
+                            ? `(${t('manage.general.pointsMultiplierDescription')} ${selectedInstanceObj.element.pointsMultiplier}x)`
```
sjschlapbach: Base points never have a multiplier applied to them, so we should probably hide this here - I understand the reasoning because these are used for the total points for all asynchronous activities, but this might definitely confuse lecturers since it is not consistent with the grading explanation

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                              {t(`shared.${instance.type}.short`)})
+                            </span>
+                            <span className="justify-left flex flex-row align-middle text-xs text-gray-500">
+                              <span className="text-violet-500">
+                                {instance.totalPoints} P.
+                              </span>
+                            </span>
+                          </span>
+                          {outdatedInstances.includes(instance.id) ? (
+                            <FontAwesomeIcon
```
sjschlapbach: If an element is outdated we could maybe also add an orange border or background or something to make it more clear / visually visible

sjschlapbach: (idea) Ideally we could even add a warning triangle on the corresponding block so that the user can quickly find the outdated question?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                                )}
+                              />
+                            </span>
+                          </span>
+                          <span className="flex flex-col">
+                            <span className="font-semibold">
+                              {instance.name} (
+                              {t(`shared.${instance.type}.short`)})
+                            </span>
+                            <span className="justify-left flex flex-row align-middle text-xs text-gray-500">
```
sjschlapbach: I would probably hide the points on the instance here, since I feel a bit like the users might get overwhelmed with the colored total points everywhere 😅

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                          )}
+                        >
+                          <span className="relative mr-2 flex h-5 w-5 items-center justify-center">
+                            <RadioGroupItem
+                              value={String(instance.id)}
+                              id={String(instance.id)}
+                              className="absolute inset-0 z-10 h-full w-full cursor-pointer opacity-0"
+                            />
+                            {/* Checkmark icon, only visible when checked */}
+                            <span className="pointer-events-none">
```
sjschlapbach: regarding consistent styling - do you think we could also use one of the uzh-blue's here?

sjschlapbach: Does it make sense to handle this with an opacity 100 or zero - thinking about future accessibility considerations, it might make more sense to only render this component if it should be shown? To make sure the content remains aligned (which I assume was the reason for the opacity-0?) we could then apply conditional styling using tailwind-merge on the parent element (some fixed margin on the left side)?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
sjschlapbach: Regarding consistent styling - until now we mainly sticked to our theme colors - i.e. primary-100 text, etc. - do you think this could also work here or is this not noticeable enough?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -57,13 +79,43 @@ function ActivityDetailsModal({
...
 
+  // State for selected radio button per stack
+  const [selectedInstance, setSelectedInstance] = useState<string>(
+    stacks[0]?.elements[0]?.id.toString() ?? ''
+  )
+
+  const selectedInstanceObj = useMemo(() => {
+    let foundInstance
+    let foundElement
+    let foundStackIndex = -1
```
sjschlapbach: Is it correct that we only need the stackIndex here to set one of the testing attributes further down? Even though it won't have a noticeable impact on performance, I think it makes the code more complex, even though we do not need it in production - maybe just remove the stack index therefore?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -57,13 +79,43 @@ function ActivityDetailsModal({
     [data?.getOutdatedElementInstances]
   )
 
+  // State for selected radio button per stack
+  const [selectedInstance, setSelectedInstance] = useState<string>(
+    stacks[0]?.elements[0]?.id.toString() ?? ''
+  )
+
+  const selectedInstanceObj = useMemo(() => {
```
sjschlapbach: I think we could significantly simplify the way the instance is obtained here by using a flatMap (to get all instances in a list) and then just search in there?

Something like the following (not verified if it does the correct thing!)
```typescript
const selectedInstanceObj = useMemo(() => {
    const instance = stacks.flatMap(s => s.elements)
      .find(i => String(i.id) === selectedInstance)
    return instance
      ? {
        instance: instance.instance as ElementInstance,
        element: instance as ActivityInfoElement,
      }
      : undefined
  }, [stacks, selectedInstance])
```

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+        <div
+          className="h-vh flex w-2/5 flex-col gap-4 overflow-x-auto overflow-y-auto pr-10"
+          data-cy="activity-details-modal"
+        >
+          <ul className="">
+            <li>
+              <span className="font-bold">
+                {t('manage.activityWizard.name')}:
+              </span>{' '}
+              {activityDetails?.activityDetails?.metadata.name}
```
sjschlapbach: nitpick: could we maybe set const details = activityDetailData?.activityDetails to avoid having these duplicated access over and over 😅

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                        }{' '}
+                        P.
+                      </ShadcnTableCell>
+                    </ShadcnTableRow>
+                  ) : null}
+                </ShadcnTableBody>
+                <ShadcnTableFooter>
+                  <ShadcnTableRow>
+                    <ShadcnTableCell colSpan={1}>
+                      {t('manage.general.totalPointsDescription')}
```
sjschlapbach: I think it would make sense to state somewhere, that this is the maximum number of achievable points, since it might otherwise seem like this is just the standard amount of points everyone can reach?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                                    ? 'opacity-100'
+                                    : 'opacity-0'
+                                )}
+                              />
+                            </span>
+                          </span>
+                          <span className="flex flex-col">
+                            <span className="font-semibold">
+                              {instance.name} (
+                              {t(`shared.${instance.type}.short`)})
```
sjschlapbach: I tend to use "template strings" for such cases - but this is just fyi / in case you are interested - no changes required :)
{`${instance.name} (${t(`shared.${instance.type}.short`})`}

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+                        </label>
+                      ))}
+                    </AccordionContent>
+                  </div>
+                </AccordionItem>
               ))}
+            </RadioGroup>
+          </Accordion>
+        </div>
+        {selectedInstanceObj && (
```
sjschlapbach: If no instance is selected, we should not only show a large white space, but ideally a notification that the user can select an element instance from any block to inspect the preview and get more information about it?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -83,90 +135,371 @@ function ActivityDetailsModal({
...
+          <div className="flex w-3/5 flex-col gap-4 overflow-y-auto pr-3">
```
sjschlapbach: I think the information if the element has a sample solution might still be missing here? Or did I miss something?

packages/graphql/src/services/activities.ts
```diff
@@ -76,6 +76,11 @@ export async function getUserActivities(ctx: ContextWithUser) {
     return null
   }
 
+  // console.log(
```
sjschlapbach: we should remove debug loggin before merging :)

packages/graphql/src/schema/activities.ts
```diff
@@ -25,13 +32,22 @@ export const ActivityInfoElement = builder.objectType(ActivityInfoElementRef, {
     id: t.exposeInt('id'),
     name: t.exposeString('name'),
     type: t.expose('type', { type: ElementType }),
+    pointsMultiplier: t.exposeInt('pointsMultiplier'),
```
sjschlapbach: Now that we fetch the entire element instance for the preview, could we remove the id, name, type, and pointsMultiplier from this level, since this is duplicate information that is fetched again through the instance?

packages/graphql/src/schema/activities.ts
```diff
@@ -178,17 +195,51 @@ export const ReducedActivityInfo = builder.objectType(ReducedActivityInfoRef, {
   }),
 })
 
+export interface IActivityDetailsMetadata {
+  name: string
+  displayName: string
+  type: string
+  pointsMultiplier: number
+  totalBasePoints: number
```
sjschlapbach: In case we decide not to pass splitted points for asynchronous activities (where only one type of points is available), this would need to be optional as well

packages/graphql/src/services/activities.ts
```diff
@@ -319,21 +324,115 @@ export async function getLiveQuizDetails(
...
+        instance.elementType === DB.ElementType.FLASHCARD
+      ) {
+        // no points awarded
+        return elementSum
+      }
+
+      const { elementData } = instance
+
+      const hasBasePoints = elementData.basePoints
+      const pointsMultiplier = elementData.pointsMultiplier
```
sjschlapbach: We need to be careful here - the content of `instance.elementData` is the data from the options of the element. Accordingly, elementData.pointsMultiplier is the multiplier of the element (only) and does not account for the multiplier on the activity that will affect the product applied on the instance. To accumulate the correct stack points we should use instance.options.pointsMultiplier instead, which is the combined one from element and activity

packages/graphql/src/services/activities.ts
```diff
@@ -319,21 +324,115 @@ export async function getLiveQuizDetails(
...
+        'hasSampleSolution' in elementData.options
+          ? ((elementData.options as { hasSampleSolution?: boolean })
+              .hasSampleSolution ?? false)
+          : false
+
+      const basePoints = isContentOrFlashcard
+        ? 0
+        : hasBasePoints
+          ? defaultPoints
+          : 0
```
sjschlapbach: Could this be written more simply as `const basePoints = !isContentOrFlashcard && hasBasePoints ? defaultPoints : 0`?

packages/graphql/src/services/activities.ts
```diff
@@ -319,21 +324,115 @@ export async function getLiveQuizDetails(
...
+        ? 0
+        : hasSampleSolution
+          ? pointsMultiplier * defaultMaxBonusPoints
+          : 0
+      const totalPoints = basePoints + (correctnessPoints + bonusPoints)
+      return {
+        id: instance.id,
+        name: instance.elementData.name,
+        type: instance.elementType,
+        pointsMultiplier: pointsMultiplier,
```
sjschlapbach: nitpick: While this should be fine through the scope of the variable, it might make sense to distinguish this from the pointsMultiplier of the activity stored in a variable with the same name (which is also available here)

packages/graphql/src/services/activities.ts
```diff
@@ -319,21 +324,115 @@ export async function getLiveQuizDetails(
...
+        sum +
+        stack.elements.reduce((elSum, el) => elSum + el.correctnessPoints, 0),
+      0
+    ),
+    totalBonusPoints: stacks.reduce(
+      (sum, stack) =>
+        sum + stack.elements.reduce((elSum, el) => elSum + el.bonusPoints, 0),
+      0
+    ),
+    totalPoints: stacks.reduce((sum, stack) => sum + stack.stackPoints, 0),
```
sjschlapbach: While probably not a performance problem, we loop over the same stacks four times here, the reduce can also return all four values at the same time and by destructuring them directly, we would also have a more compact representation :)

Something like:
```typescript
const {base, correctness, bonus, total} = stacks.reduce(
    (acc, stack) => {
      for (const el of stack.elements) {
        acc.base += el.basePoints
        acc.correctness += el.correctnessPoints ?? 0
        acc.bonus += el.bonusPoints ?? 0
      }
      acc.total += stack.stackPoints
      return acc
    },
    { base: 0, correctness: 0, bonus: 0, total: 0 }
  )
```

packages/graphql/src/services/activities.ts
```diff
@@ -354,20 +453,75 @@ export async function getPracticeQuizDetails(
...
       : 0,
-    elements: block.elements.map((instance) => ({
-      id: instance.id,
-      name: instance.elementData.name,
-      type: instance.elementType,
-    })),
+    stackPoints: block.elements.reduce((elementSum, instance) => {
+      if (
+        instance.elementType === DB.ElementType.CONTENT ||
+        instance.elementType === DB.ElementType.FLASHCARD
```
sjschlapbach: Could we also return early for asynchronous activities, if we do not have a sample solution defined?

sjschlapbach (changes requested):
All the required functionalities are already there, and the main idea looks really good! I added some comments where I think minor bugs might have slipped in and various other ideas / nitpicking details to follow the codestyle of the remaining codebase.
Additionally, this PR would probably also benefit strongly from cypress coverage to ensure complete functionality

packages/prisma/src/data/seedTEST.ts
jabbadizzleCode: "No 'User' record(s) (needed to inline the relation on 'AnswerCollection' record(s)) was found for a nested connect on one-to-many relation 'AnswerCollectionToUser'."

sjschlapbach: @jabbadizzleCode sorry for the confusion, this was solved with another PR merged into the `v3` branch a few days ago - the USER_ID_TEST should be imported directly from the main file where the ids are defined. Merging the v3 back into this branch and resolving the conflict accordingly should result in the desired behavior :)

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
jabbadizzleCode: So do we only need to display the basePoints as e.g. "Points (Multiplier X)" and no totalPoints? Or just totalPoints with the multiplier?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
jabbadizzleCode: So should I also hide the multiplier for asynchronous activities?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
jabbadizzleCode: Not needed for cypress?

packages/graphql/src/services/activities.ts
jabbadizzleCode: If there is no sample solution, should it return 0 points?

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -84,89 +144,367 @@ function ActivityDetailsModal({
...
-                        'hover:text-primary-100 flex flex-row items-center justify-between gap-1.5 border-b text-sm',
-                        outdatedInstances.includes(instance.id)
-                          ? 'bg-uzh-red-20'
-                          : ''
-                      )}
-                      data-cy={`stack-${index}-instance-${instanceIx}`}
-                    >
-                      <div className="flex flex-row items-center gap-1.5">
-                        {outdatedInstances.includes(instance.id) ? (
+                  <div key={stack.id} className="w-full">
```
sjschlapbach: We should remove the key (at least - if not the entire div? - here, to avoid duplications with the element above

apps/frontend-manage/src/components/activities/overview/ActivityDetailsModal.tsx
```diff
@@ -84,89 +144,367 @@ function ActivityDetailsModal({
...
+                        {activity.type === ActivityType.LiveQuiz
+                          ? t('shared.generic.blockN', {
+                              number: index + 1,
+                            })
+                          : t('shared.generic.stackN', {
+                              number: index + 1,
+                            })}
+                      </span>
+                      <div className="flex flex-row items-center justify-end gap-3">
+                        {stack.elements.filter((element) =>
```
sjschlapbach: idea: this can also be done by checking for stack.elements.some((element =< outdatedInstances.includes(element.instance.id))

PR #4815, opened 2025-08-01
fix(apps/olat-api): make sure that permissions on course are correctly validated during OLAT API request

sjschlapbach:
## Summary by CodeRabbit

* **New Features**
  * Requests to course-specific account creation pages now automatically redirect to a centralized account creation page.

* **Bug Fixes**
  * Improved course and activity retrieval to ensure users with appropriate permissions can access shared, non-archived courses and related activities, not just those they own.

* **Chores**
  * Updated backend dependencies for improved utility management.
  * Enhanced permission handling by centralizing permission recalculations during test data creation for better consistency and clarity.
  * Added build step for utility package in test environment setup to ensure proper dependency compilation.

### DIFF
--- a/apps/olat-api/src/services.ts
+++ b/apps/olat-api/src/services.ts
@@ -24,19 +24,20 @@ export async function getCourses(provider: string, providerAccountId: string) {
     select: {
       user: {
         select: {
-          courses: {
-            where: { isArchived: false },
-            select: { id: true, name: true },
+          // find all shared non-archived courses (permission level irrelevant - read permissions are sufficient)
+          objects: {
+            where: { courseId: { not: null }, course: { isArchived: false } },
+            select: { course: { select: { id: true, name: true } } },
           },
         },
       },
     },
   })
 
-  const courses = account?.user?.courses ?? []
-  if (courses.length === 0) {
-    return []
-  }
+  const courses =
+    account?.user?.objects
+      .map((object) => object.course)
+      .filter((course) => !!course) ?? []
 
   return courses.map((course) => ({
     id: course.id,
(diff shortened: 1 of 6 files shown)


apps/olat-api/src/services.ts
```diff
@@ -24,19 +24,20 @@ export async function getCourses(provider: string, providerAccountId: string) {
     select: {
       user: {
         select: {
-          courses: {
-            where: { isArchived: false },
-            select: { id: true, name: true },
+          // find all shared non-archived courses (permission level irrelevant - read permissions are sufficient)
+          objects: {
+            where: { courseId: { not: null }, course: { isArchived: false } },
+            select: { course: { select: { id: true, name: true } } },
```
sjschlapbach, line 30: @jabbadizzleCode Through the "objects" on the user we can directly join all objects of a type (going through the derivedPermissions table). With the filters added here, we only include courses that have the archived boolean set to false

apps/olat-api/src/services.ts
```diff
@@ -85,14 +86,15 @@ export async function getCourseActivityTypes(
...
-    },
+    select: { userId: true },
   })
   if (!account) return null
 
   const course = await prisma.course.findUnique({
-    where: { id: courseID, ownerId: account.userId },
+    where: {
+      id: courseID,
+      permissions: { some: { userId: account.userId } }, // user has at least read permissions on course
```
sjschlapbach, line 96: @jabbadizzleCode  To validate the access of an individual user to an object, we check for the corresponding **derived permission** entry (usually also with a permission level - since the lowest permission level read is sufficient here, this can be left out). Note the difference between "permissions", which are "DerivedPermissions" and "directPermission", which are linked to the "Permission" table, representing directly granted permissions (sharing actions / specific import functionalities)

jabbadizzleCode (commented):
Looks good to me :)

PR #4839, opened 2025-08-15
enhance: add hatchet infrastructure (initial deployments for hatchet-ha, hatchet-lite for local development), add initial response-api and response-processor worker based on hatchet

rschlaefli:
## Summary by CodeRabbit

- **New Features**
  - Public Response API (/AddResponse) with health/CORS and optional dual-submit to v1+v2.
  - Two Hatchet workers plus shared tasks package and local Hatchet service with dev token helper.
  - Frontend toggle and v2 endpoint config to enable dual-response mode.
  - Redis-backed response processor: scoring, XP, leaderboards, duplicate protection, and optional dual-namespace parity mode.

- **Chores**
  - CI/workflow additions for multi-arch image builds, Docker Compose + Helm charts/values for Hatchet and response-api, DB init, and Redis comparison/reporting tooling.


deploy/charts/klicker-uzh-v2/templates/cm-hatchet-workers.yaml
sjschlapbach, line 1: While deploying a new version to our staging environment, I noticed that the pods do not seem to be replaced, but new versions seem to be created, with the old ones remaining on the cluster with the error `ImagePullBackOff`. Could this have something to do with the configuration here?

apps/hatchet-worker-general/src/hatchet-client.ts
```diff
@@ -0,0 +1,9 @@
+import { HatchetClient } from '@hatchet-dev/typescript-sdk'
+
+export const hatchet = HatchetClient.init({
+  token: process.env.HATCHET_CLIENT_TOKEN,
+  api_url: process.env.HATCHET_API_URL,
+  tenant_id: process.env.HATCHET_TENANT_ID,
+  host_port: process.env.HATCHET_HOST_PORT,
+  log_level: 'DEBUG',
```
sjschlapbach: Should we maybe set this value based on the NODE_ENV?

apps/hatchet-worker-general/src/index.ts
```diff
@@ -0,0 +1,16 @@
+import { prepareHatchetTasks } from '@klicker-uzh/hatchet-tasks'
+import { hatchet } from './hatchet-client.js'
+
+async function main() {
+  const workflows = prepareHatchetTasks(hatchet)
+
+  const worker = await hatchet.worker('hatchet-worker-general', {
+    // TODO: figure out some way to share tasks with backend (e.g., try shared tasks package)
```
sjschlapbach, line 8: Is this TODO still up to date? On the v3-hatchet branch we managed to already do this :)

apps/hatchet-worker-response-processor/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const MAX_BONUS_POINTS = 45 // maximum 45 bonus points for fastest answer
```
sjschlapbach, line 1: Could we possibly import these constants from the shared-components package when we already rewrite the implementation - to only have a single place where these values are defined?

apps/hatchet-worker-response-processor/src/hatchet-client.ts
```diff
@@ -0,0 +1,9 @@
+import { HatchetClient } from '@hatchet-dev/typescript-sdk'
+
+export const hatchet = HatchetClient.init({
+  token: process.env.HATCHET_CLIENT_TOKEN,
+  api_url: process.env.HATCHET_API_URL,
+  tenant_id: process.env.HATCHET_TENANT_ID,
+  host_port: process.env.HATCHET_HOST_PORT,
+  log_level: 'DEBUG',
```
sjschlapbach: Same as above - maybe we could set the log_level based on the NODE_ENV?

packages/hatchet-tasks/rollup.config.js
sjschlapbach, line 1: Does it make sense to handle this through a separate package? On the `v3-hatchet` branch I implemented the corresponding functions as parts of the backend services functions and then only imported them in the `apps/hatchet` package, for which a separate package is not really required?
This makes sense in my opinion, since we quite frequently use similar / identical code as in other backend functions, which would otherwise be duplicated with a hatchet tasks package?

sjschlapbach (commented):
LGTM - added some comments for my understanding and I think it might make sense to review some of the Coderabbit comments, since they seemed to make sense. Additionally, we should probably verify that we really only deploy the components in production where we are sure that we want them - some of the production values files had replicaCounts different from zero (not sure if the relevant ones though)

While re-deploying to the staging environment, I noticed that the hatched pods are not replaced, but new ones are added and the old ones fail. Maybe we could fix this before merging the branch? :)

Additionally, what would be the required setup to run the functionality locally for development? Could we maybe add corresponding .env.template files if required or a short description of the setup how to run it?

packages/hatchet-tasks/rollup.config.js
rschlaefli: will revisit this when merging into v3-hatchet and maybe change. for now, will leave it like this as it is a workable example (and did not want to create conflicts with graphql package).

apps/hatchet-worker-general/src/hatchet-client.ts
rschlaefli: probably better to use a separate env variable. will add.

apps/hatchet-worker-response-processor/src/constants.ts
rschlaefli: will have to figure out a way for storing such constants once across different runtimes (e.g., once we have TS azure function + go container response-api). might be easier to just do this in the environment variables, or with some config file that is loaded by both and built into the containers. so shared-components is probably not the right place, as it is config and not really TS application parts.

apps/hatchet-worker-response-processor/src/hatchet-client.ts
rschlaefli: dito

apps/hatchet-worker-response-processor/src/constants.ts
sjschlapbach: Yes, this makes sense - if this is not too difficult to do, this sounds like a significantly better solution :)

PR #4854, opened 2025-08-23
refactor: rework prisma package for workable integration with auth and apps

rschlaefli:


packages/prisma/src/prisma/schema/js.prisma
```diff
@@ -1,15 +1,18 @@
 generator client {
-  provider        = "prisma-client-js"
-  previewFeatures = ["postgresqlExtensions", "views", "queryCompiler", "driverAdapters"]
-  output          = "../client"
-  binaryTargets   = ["native", "debian-openssl-1.1.x", "linux-musl-openssl-3.0.x", "debian-openssl-3.0.x"]
+  provider               = "prisma-client"
+  previewFeatures        = ["postgresqlExtensions", "views", "queryCompiler", "driverAdapters"]
+  output                 = "../client"
+  runtime                = "nodejs"
+  moduleFormat           = "esm"
```
rschlaefli, line 6: prisma client will now be built to .ts files as a clean ES module and then will be built into our own "client" with rollup

packages/prisma/src/prisma/schema/js.prisma
```diff
@@ -1,15 +1,18 @@
...
+  generatedFileExtension = "ts"
+  importFileExtension    = "js"
 }
 
-generator pothos {
-  provider     = "prisma-pothos-types"
-  clientOutput = "@klicker-uzh/prisma"
-  output       = "../client/pothos.ts"
-}
+// generator pothos {
```
rschlaefli: we did not actually ever use this, so removed for now

packages/prisma/src/index.ts
```diff
@@ -0,0 +1,27 @@
+import { PrismaPg } from '@prisma/adapter-pg'
+import { PrismaClient } from './client.js'
+
+// TODO: figure out whether using Pool with pg is a good idea for us (or does pgbouncer do that server-side)
```
rschlaefli, line 4: afaik the connection pooling is done by pgbouncer on azure, but we could still investigate whether this would be useful

packages/prisma/src/index.ts
```diff
@@ -0,0 +1,27 @@
+import { PrismaPg } from '@prisma/adapter-pg'
```
rschlaefli, line 1: master prisma client setup file -> one place to optimize and import everywhere instead of installing it in all places and doing the same thing repeatedly

sjschlapbach (approved):
LGTM - also tested it and everything seems to be working as before :)
Only thing we might have to be careful with / verify before merging it into the v3 is that the database migration does not result in issues depending on the creation order in connection with other PRs we might merge before?

PR #4863, opened 2025-08-27
feat(apps/chat): add new chatbot frontend 

TheHummel:
## Summary by CodeRabbit

- New Features
  - Introduced a new Chat app with authenticated access, threaded conversations (create/rename/delete), message editing/reloading, and branch navigation.
  - Model and mode selection with fallback options; real-time streaming replies; credits display and automatic deduction.
  - Rich message rendering (Markdown, math), code copy, and tool UI (RAG search).
  - Mobile-friendly UI and sidebar with history and settings.

- Documentation
  - Added chat README guidance and env examples.

- Chores
  - Added PWA redirect to Chat, environment variables across apps, and Traefik routes.
  - New Dockerfile, QA build workflow, and Helm/Kubernetes templates for deploying the Chat service.
  - Updated dependencies and build config.


apps/chat/app/api/chat/route.ts
```diff
@@ -0,0 +1,181 @@
...
+  convertToModelMessages,
+  stepCountIs,
+  streamText,
+  UIMessage,
+  type LanguageModel,
+} from 'ai'
+
+export const maxDuration = 30
+
+const prisma = new PrismaClient()
```
rschlaefli: pull in v3 and update this to the new prisma approach wherever prisma is used (now we can import { prisma } and just use it without creating a new client. and { PrismaClient } if needed has to be imported from @klicker-uzh/prisma/client

apps/chat/app/api/chat/route.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export async function POST(req: Request) {
+  const {
+    messages,
+    threadId,
+    selectedModel,
+    systemPrompt,
```
rschlaefli: the system prompt should not come from the frontend/client (as then it could be easily overwritten by smart people). instead, send the "chatMode" (tutor, explainer) back and use a prompts file/dict to store the prompts (for now).

apps/chat/app/api/threads/[threadId]/messages/route.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const messages = await prisma.chatMessage.findMany({
+      where: { threadId },
+      orderBy: { createdAt: 'asc' },
+    })
+
+    return Response.json(
+      messages.map((msg) => ({
+        id: msg.id,
+        thread_id: msg.threadId,
```
Comment (rschlaefli) [inline apps/chat/app/api/threads/[threadId]/messages/route.ts]: why are we mixing thread_id and threadId casing here?

apps/chat/app/api/threads/[threadId]/messages/route.ts
```diff
@@ -0,0 +1,70 @@
...
+        updated_at: msg.updatedAt.toISOString(),
+      }))
+    )
+  } catch (error) {
+    console.error('Failed to fetch messages:', error)
+    return Response.json({ error: 'Failed to fetch messages' }, { status: 500 })
+  }
+}
+
+export async function POST(
```
Comment (rschlaefli) [inline apps/chat/app/api/threads/[threadId]/messages/route.ts]: not sure I understand what this code does (it looks very similar to the code in the chat route) -> please add a short jsdoc comment to these endpoints for clarification

apps/chat/app/api/threads/[threadId]/title/route.ts
```diff
@@ -0,0 +1,26 @@
+import { PrismaClient } from '@klicker-uzh/prisma'
+
+const prisma = new PrismaClient()
+
+export async function PUT(
```
Comment (rschlaefli) [inline apps/chat/app/api/threads/[threadId]/title/route.ts]: not sure if we actually need such a feature (thread titles would be ai generated in the future in the optimal case)

apps/chat/app/api/threads/[threadId]/route.ts
```diff
@@ -0,0 +1,54 @@
+import { PrismaClient } from '@klicker-uzh/prisma'
+
+const prisma = new PrismaClient()
+
+export async function GET(
```
Comment (rschlaefli) [inline apps/chat/app/api/threads/[threadId]/route.ts]: there seem to be a lot of endpoints that look very similar and do similar mappings -> can we check whether all of them are needed and next to documentation, we should think about extracting the logic into a thread.ts service

apps/chat/app/api/threads/[threadId]/route.ts
```diff
@@ -0,0 +1,54 @@
...
+      updated_at: thread.updatedAt.toISOString(),
+      message_count: thread._count.messages,
+    })
+  } catch (error) {
+    console.error('Failed to fetch thread:', error)
+    return Response.json({ error: 'Failed to fetch thread' }, { status: 500 })
+  }
+}
+
+export async function DELETE(
```
Comment (rschlaefli) [inline apps/chat/app/api/threads/[threadId]/route.ts]: for all cases, we will need to to authorization next to authentication to make sure that only the owner of a thread can delete it -> this will require adjusting all of the prisma interactions to have a where ownerId

apps/chat/app/api/threads/route.ts
```diff
@@ -0,0 +1,52 @@
...
+    console.error('Failed to fetch threads:', error)
+    return Response.json([], { status: 500 })
+  }
+}
+
+export async function POST(req: Request) {
+  try {
+    const { title } = await req.json()
+
+    const thread = await prisma.chatThread.create({
```
rschlaefli: another function that creates a new thread -> with a threads.ts service the logic could be probably reused?

apps/chat/app/stores/chatStore.ts
```diff
@@ -0,0 +1,543 @@
+'use client'
```
rschlaefli: this file is quite critical (I guess) but quite hard to understand -> please split the util function from the store logic and add more documentation for the functions

apps/chat/app/stores/settingsStore.ts
```diff
@@ -0,0 +1,110 @@
...
+          description: 'OpenAI model',
+        },
+        {
+          id: 'anthropic',
+          name: 'Claude 4 Sonnet',
+          description: 'Anthropic model',
+        },
+      ],
+
+      modeOptions: [
```
rschlaefli: maybe we could actually just extract this to a separate constant map and then pass the mode to the api and get the prompt from here. the prompts will be very very long so it would be unreadable if we added them to this file anyway

apps/chat/app/globals.css
```diff
@@ -0,0 +1,122 @@
+@import 'tailwindcss';
```
rschlaefli: not sure whether this matches the config for our design system already (including source sans 3 etc as font and uzh colors), check again with the other klicker apps

apps/chat/app/layout.tsx
```diff
@@ -0,0 +1,35 @@
+import 'katex/dist/katex.min.css'
+import type { Metadata } from 'next'
+import { Geist, Geist_Mono } from 'next/font/google'
+import './globals.css'
+
+const geistSans = Geist({
```
rschlaefli: change to our uzh font source sans 3. keep the mono font as is

apps/chat/app/layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+  subsets: ['latin'],
+})
+
+const geistMono = Geist_Mono({
+  variable: '--font-geist-mono',
+  subsets: ['latin'],
+})
+
+export const metadata: Metadata = {
+  title: 'assistant-ui Starter App',
```
rschlaefli:
```suggestion
  title: 'KlickerUZH Tutor Chatbot',
```

apps/chat/app/layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+  description: 'Generated by create-assistant-ui',
```
rschlaefli: Remove the description for now

apps/chat/app/RuntimeProvider.tsx
```diff
@@ -0,0 +1,492 @@
...
+  // AbortController to handle request cancellation
+  const abortControllerRef = useRef<AbortController | null>(null)
+
+  // load threads on mount
+  useEffect(() => {
+    useChatStore.getState().loadThreads()
+  }, [])
+
+  // function to handle streaming chat response
+  const generateChatResponse = useCallback(
```
rschlaefli: can we extract these long callbacks/hooks into separate files? they make this component very long and are not so easy to understand (might profit from improved structure/comments)

apps/chat/app/RuntimeProvider.tsx
```diff
@@ -0,0 +1,492 @@
...
+            const chunk = decoder.decode(value, { stream: true })
+            buffer += chunk
+
+            const lines = buffer.split('\n')
+            buffer = lines.pop() || ''
+
+            for (const line of lines) {
+              if (!line.trim()) continue
+
+              if (line === 'data: [DONE]') {
```
rschlaefli: these things seem critical but it would be very helpful to explain them in a function comment (or to explain the main idea somewhere else)

apps/chat/app/tools.ts
```diff
@@ -0,0 +1,173 @@
...
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ query, top_k }),
+      }
+    )
+    const result = await response.json()
+    return result
+  },
+})
+
+export const context7 = tool({
```
rschlaefli: this does not make sense -> it is an MCP and we should really test whether the MCP tool functionality of the AI SDK works (then we would no code at all to integrate it). right now it is basically using the MCP as an API, which is not what corresponds to the idea of MCPs

apps/chat/components/assistant-ui/tools-ui/context7-tool-ui.tsx
```diff
@@ -0,0 +1,213 @@
+'use client'
```
rschlaefli: cool example of a custom UI but do not spend more time here as the context7 MCP was thought of as an example to test the MCP integration. students will not actually need context7.

apps/chat/components/assistant-ui/markdown-text.tsx
```diff
@@ -0,0 +1,249 @@
+'use client'
+
+import '@assistant-ui/react-markdown/styles/dot.css'
+import rehypeKatex from 'rehype-katex'
```
rschlaefli: I agree that maybe katex css is needed? we had to import it in klicker to nicely display latex

apps/chat/components/assistant-ui/markdown-text.tsx
```diff
@@ -0,0 +1,249 @@
...
+import { TooltipIconButton } from '@/components/assistant-ui/tooltip-icon-button'
+import { cn } from '@/lib/utils'
+
+const MarkdownTextImpl = () => {
+  return (
+    <MarkdownTextPrimitive
+      remarkPlugins={[remarkGfm, remarkMath]}
+      rehypePlugins={[rehypeKatex]}
+      preprocess={normalizeCustomMathTags}
+      className="aui-md"
```
rschlaefli: sanitize would be in fact a good idea, we have a config in our markdown component that might work

apps/chat/components/assistant-ui/settings-panel.tsx
```diff
@@ -0,0 +1,116 @@
...
+      <div className="space-y-2">
+        <label className="text-muted-foreground text-xs">Chat Mode</label>
+        <div className="flex items-center justify-between">
+          <div className="flex items-center space-x-3">
+            <span
+              className={`text-xs ${selectedMode === 'tutor' ? 'font-medium' : 'text-muted-foreground'}`}
+            >
+              Tutor
+            </span>
+            <Switch
```
rschlaefli: this makes more sense as a dropdown rather than a switch. lecturers will in the future be able to customize the modes they offer. so they might have only one (tutor or explainer), both, or more than two. we should also already read all of this information from a constants file

apps/chat/components/assistant-ui/thread.tsx
```diff
@@ -0,0 +1,260 @@
...
+        <div className="flex w-full flex-grow flex-col items-center justify-center">
+          <p className="mt-4 font-medium">How can I help you today?</p>
+        </div>
+        <ThreadWelcomeSuggestions />
+      </div>
+    </ThreadPrimitive.Empty>
+  )
+}
+
+const ThreadWelcomeSuggestions: FC = () => {
```
rschlaefli: are these actually shown to the user? if yes, we should think about how we can make these more generic to match the course chatbot use case

apps/chat/components/ui/breadcrumb.tsx
```diff
@@ -0,0 +1,109 @@
+import { Slot } from '@radix-ui/react-slot'
```
rschlaefli: for each ui component here, try whether it also works with its design system counterpart. if yes or with minimal customization, remove it from here to reduce code duplication

apps/chat/next.config.ts
```diff
@@ -0,0 +1,7 @@
+import type { NextConfig } from 'next'
+
+const nextConfig: NextConfig = {
+  /* config options here */
+}
+
```
rschlaefli, line 28: agree with coderabbit

apps/frontend-control/.env.production
```diff
@@ -10,6 +10,7 @@ NEXT_PUBLIC_MANAGE_URL="https://manage.klicker.uzh.ch"
 NEXT_PUBLIC_CONTROL_URL="https://control.klicker.uzh.ch"
 NEXT_PUBLIC_AUTH_URL="https://auth.klicker.uzh.ch"
 NEXT_PUBLIC_LTI_URL="https://lti.klicker.uzh.ch"
+NEXT_PUBLIC_CHAT_URL="https://chat.klicker.com"
```
rschlaefli: in the production files, it is klicker.uzh.ch

apps/frontend-control/.env.qa
```diff
@@ -10,3 +10,4 @@ NEXT_PUBLIC_MANAGE_URL="https://manage.klicker-qa.bf-app.ch"
 NEXT_PUBLIC_CONTROL_URL="https://control.klicker-qa.bf-app.ch"
 NEXT_PUBLIC_AUTH_URL="https://auth.klicker-qa.bf-app.ch"
 NEXT_PUBLIC_LTI_URL="https://lti.klicker-qa.bf-app.ch"
+NEXT_PUBLIC_CHAT_URL="https://chat.klicker.com"
```
rschlaefli: coderabbit is right, klicker-qa.bf-app.ch is the QA domain

apps/frontend-manage/.env.production
```diff
@@ -10,6 +10,7 @@ NEXT_PUBLIC_MANAGE_URL="https://manage.klicker.uzh.ch"
 NEXT_PUBLIC_CONTROL_URL="https://control.klicker.uzh.ch"
 NEXT_PUBLIC_AUTH_URL="https://auth.klicker.uzh.ch"
 NEXT_PUBLIC_LTI_URL="https://lti.klicker.uzh.ch"
+NEXT_PUBLIC_CHAT_URL="https://chat.klicker.com"
```
rschlaefli: dito

apps/frontend-pwa/.env.production
```diff
@@ -11,6 +11,7 @@ NEXT_PUBLIC_WEB_PUSH_PUBLIC_KEY="BPz0NNgACtulol5iYwmnFzzL-gkTUQ2jZrXsdrr9oEfhZUB
 NEXT_PUBLIC_PWA_URL="https://pwa.klicker.uzh.ch"
 NEXT_PUBLIC_MANAGE_URL="https://manage.klicker.uzh.ch"
 NEXT_PUBLIC_CONTROL_URL="https://control.klicker.uzh.ch"
+NEXT_PUBLIC_CHAT_URL="https://chat.klicker.com"
```
rschlaefli: dito

packages/prisma/src/prisma/schema/chat.prisma
```diff
@@ -0,0 +1,26 @@
+model ChatThread {
```
rschlaefli, line 19: we need to think about how we link this with the participant/their account. will get back to you once eduid for students is working.

rschlaefli (changes requested):
Already looks quite good :) Reviewed the code without running the UI right now, so let's look at the UI together soon. Please look at all of the Coderabbit comments as well, as many are quite smart.

apps/chat/app/api/chat/route.ts
TheHummel: done

apps/chat/app/api/threads/[threadId]/messages/route.ts
Comment (TheHummel) [reply apps/chat/app/api/threads/[threadId]/messages/route.ts]: removed as it wasn't needed

apps/chat/app/api/threads/[threadId]/title/route.ts
Comment (TheHummel) [reply apps/chat/app/api/threads/[threadId]/title/route.ts]: I agree

apps/chat/app/tools.ts
TheHummel: replaced it with ai sdk's create_mcpclient interface and respective tools attribute now

apps/chat/components/assistant-ui/thread.tsx
TheHummel: yes these are the default suggestions form assistant-ui. maybe one option might be to also have lecturers create them or instead have them automatically generated based on the script once a new chatbot instance is created by a lecturer?

packages/prisma/src/prisma/schema/migrations/20250910125127_make_chatbot_required_in_threads/migration.sql
```diff
@@ -0,0 +1,101 @@
+/*
```
rschlaefli: At the end of a PR, can delete all migrations and recreate them as one single clean migration (then there is not so much stuff added and then dropped again)

apps/chat/src/lib/config/suggestions.ts
```diff
@@ -0,0 +1,27 @@
...
+    id: 'suggestion-1',
+    text: 'Explain a random concept from the script',
+    prompt:
+      'Take a random concept from the course script and explain it in simple terms.',
+  },
+  {
+    id: 'suggestion-2',
+    text: 'Help me prepare for the exam',
+    prompt:
+      'Create a study plan for the upcoming exam covering all key topics based on the lecture materials.',
```
rschlaefli, line 18: This is probably not a realistic question as the chatbot does not have the "big picture"

apps/chat/src/lib/config/models.ts
```diff
@@ -0,0 +1,48 @@
...
+  {
+    id: 'gpt-4.1',
+    name: 'GPT-4.1',
+    description: 'OpenAI model',
+    cost: {
+      input: 0.25,
+      output: 2.0,
+    },
+  },
+  {
```
rschlaefli, line 28: Remove sonnet (too expensive), add gpt-4.1-mini as a fallback when no credits anymore

apps/chat/src/lib/api/utils.ts
```diff
@@ -0,0 +1,171 @@
+import { ExtendedThreadMessageLike } from '../../stores/chatStore'
+
+/**
+ * Generates unique thread ID with timestamp and random suffix
+ * @returns A unique thread identifier string
+ */
+export const generateThreadId = (): string =>
+  `thread-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`
```
rschlaefli: We should use uuid here too, or is the format of the id critical?

apps/chat/app/api/chat/route.ts
```diff
@@ -0,0 +1,140 @@
+import { getWeather, RAGSearch } from '@/app/tools'
+import { anthropic } from '@ai-sdk/anthropic'
+import { openai } from '@ai-sdk/openai'
+import { PrismaClient } from '@klicker-uzh/prisma'
```
rschlaefli: switch to the new import { prisma } from '@klicker-uzh/prisma' -> you get a working client

apps/chat/src/lib/config/suggestions.ts
TheHummel: shall I comment out the suggestions for now or delete entirely?

apps/chat/src/lib/api/utils.ts
TheHummel: yes, seems to be leftover function that was no longer used. deleted now

apps/chat/app/api/chat/route.ts
TheHummel: seems to be old version maybe. I think the {prisma} is used in all routes now :)

apps/chat/.env.development
```diff
@@ -0,0 +1,16 @@
...
+# Next.js API Config
+NEXT_PUBLIC_API_URL=http://localhost:8000
+
+# Backend API URL for client-side requests
+NEXT_PUBLIC_BACKEND_URL=http://localhost:8000
+
+# MCP Config
+MCP_URL=https://mcp-studies.stg.env.df-app.ch/mcp/mcp
+
+AZURE_RESOURCE_NAME=klicker-ai
```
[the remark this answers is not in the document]
TheHummel: @rschlaefli not sure about this. where should this var be saved ideally? this one is used for ai sdk to build the correct url

PR #4872, opened 2025-09-04
feat: add audit log for assessment mode

rschlaefli:
## Summary by CodeRabbit

* **New Features**
  * Added a standalone Audit Service with HTTP endpoints, readiness & metrics, and integrated audit logging across backend, workers, auth, and frontend assessment flows.
* **Chores/Infrastructure**
  * Added Azurite (Azure Storage emulator) to local/CI, startup readiness checks, Docker/Helm manifests, env examples, and release workflows; standardized Postgres test images.
* **Tests**
  * Added extensive API, integration, performance, and scenario test suites and CI jobs to build, test, and publish the Audit Service.

## ClickUp Links

> *This was generated by AI during triage.*

- Primary task: [MeF audit logging and pure-infra semester-start readiness](https://app.clickup.com/t/86caazaea)


.github/workflows/cypress-testing.yml
```diff
@@ -330,6 +330,19 @@ jobs:
           SERVER_LOGGER_LEVEL: debug
           SERVER_LOGGER_FORMAT: console
 
+      azurite:
```
rschlaefli, line 338: Azurite is a local emulator for blob/table/queue storage. We could probably start it only in assessment mode though (maybe another flag for the run app dependencies?)

deploy/env-prod-v3/values.yaml
```diff
@@ -533,6 +533,29 @@ responseApi:
...
+      cpu: 500m
+      memory: 512Mi
+    limits:
+      cpu: 1000m
+      memory: 1Gi
+
+  azureStorage:
+    tableName: 'auditlogs-prod'
+
+  corsOrigins: 'https://assessment.klicker.uzh.ch'
```
rschlaefli, line 557: do we ever need any frontend audit logging from manage?

.github/workflows/cypress-testing.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         containers: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
     services:
       postgres:
-        image: postgres:15
+        image: postgres:14.18
```
rschlaefli, line 47: made this actually aligned with what we have on azure

.github/workflows/v3_audit-prod.yml
```diff
@@ -0,0 +1,47 @@
+name: Build Docker image for audit (Production)
```
rschlaefli, line 1: @jabbadizzleCode will require new pipelines on the your branch as well

apps/audit/src/middleware/auth.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { Context, Next } from 'hono'
+import { config } from '../config.js'
+import { logger } from '../utils/logger.js'
+
+export interface ParticipantContext {
+  participantId: string
+  role: string
+}
+
+export async function authMiddleware(
```
rschlaefli, line 11: backend needs to send a fixed token in the header. frontend sends the student login JWT. could probably merge to both being a JWT with different issuers and according validation, but for now this will be enough

apps/audit/src/routes/audit-private.ts
```diff
@@ -0,0 +1,126 @@
...
+): void {
+  app.post(
+    '/audit',
+    authMiddleware,
+    zValidator('json', AuditEventSchema),
+    async (c) => {
+      const startTime = Date.now()
+      const event = c.req.valid('json')
+
+      metrics.requestsTotal.inc()
```
rschlaefli, line 26: idea: expose certain metrics for "prometheus", which will run on the new infra and can monitor apps. it could scale up the replicas if e.g. we have a lot of total requests, or many more requests than writes, etc.

if this works well, should expand on this in our own backend too, so we dont have only CPU/GPU usage metrics

apps/audit/src/routes/audit-private.ts
```diff
@@ -0,0 +1,126 @@
...
+import { createAuditEntity } from '../storage/entities.js'
+import type { AuditTableClient } from '../storage/table-client.js'
+import { logger } from '../utils/logger.js'
+import { metrics } from '../utils/metrics.js'
+
+interface Dependencies {
+  tableClient: AuditTableClient
+}
+
+export function registerAuditPrivateRoutes(
```
rschlaefli, line 14: the actual endpoints are very simple on purpose. we dont reject events as long as they are properly authenticated (not a lot of validation on e.g. allowed types beside the public endpoint). to make sure that w log something invalid rather than dropping it with an error message

apps/audit/src/routes/audit-public.ts
```diff
@@ -0,0 +1,252 @@
...
+      if (!participant) {
+        logger.warn(
+          { requestId: c.get('requestId'), path: c.req.path },
+          'Public endpoint authentication failed: invalid or expired participant token'
+        )
+        return c.json({ error: 'Invalid or expired participant token' }, 401)
+      }
+
+      const rateLimitKey = `public:${participant.participantId}`
+      if (!checkRateLimit(rateLimitKey, 100, 60000)) {
```
rschlaefli, line 63: we have to tune this, but a given participant should not be able to send too many events to the public endpoint (just for abuse prevention). maybe 100 in one minute is too low though. we could also have rate limits on the public endpoint before validation (e.g., based on IP).

apps/audit/src/routes/audit-public.ts
```diff
@@ -0,0 +1,252 @@
...
+            error: error instanceof Error ? error.message : 'Unknown error',
+          },
+          'Public endpoint received invalid JSON payload'
+        )
+        return c.json({ error: 'Invalid JSON body' }, 400)
+      }
+
+      const {
+        attributes,
+        subject: _subjectIgnored,
```
rschlaefli, line 101: in the public case, the subject is always the user that sends to the public endpoint, so we ignore what they might pass along in the request

apps/audit/src/routes/audit-public.ts
```diff
@@ -0,0 +1,252 @@
...
+        scope: _scopeIgnored,
+        schemaVersion: _schemaVersionIgnored,
+        ...rest
+      } = (rawBody ?? {}) as Record<string, unknown>
+
+      const attributeRecord =
```
rschlaefli, line 107: attributes could be whatever is passed in the request right now, we might need to restrict this in the future. zod does some validation on the max size of it already (as table storage caps it as well)

apps/audit/src/schemas/audit-event.ts
```diff
@@ -0,0 +1,119 @@
...
+import { z } from 'zod'
+
+export const ALLOWED_PUBLIC_ACTIONS = [
+  AuditAction.PARTICIPANT_VIEW_INSTANCE,
+  AuditAction.PARTICIPANT_SUBMIT_RESPONSE,
+  AuditAction.PARTICIPANT_UPDATE_ANSWER,
+  AuditAction.PARTICIPANT_JOIN_QUIZ,
+  AuditAction.PARTICIPANT_QUIZ_PIN_SUCCESS,
+  AuditAction.PARTICIPANT_QUIZ_PIN_FAILED,
+  AuditAction.CLIENT_ERROR,
```
rschlaefli, line 15: any error on the client caught by an error boundary

apps/audit/src/schemas/audit-event.ts
```diff
@@ -0,0 +1,119 @@
+import { AuditAction, AuditScope } from '@klicker-uzh/types'
+import { z } from 'zod'
+
+export const ALLOWED_PUBLIC_ACTIONS = [
```
rschlaefli, line 8: implementation in the UI is WIP on the other branch

apps/audit/src/schemas/audit-event.ts
```diff
@@ -0,0 +1,119 @@
...
+  AuditAction.PARTICIPANT_UPDATE_ANSWER,
+  AuditAction.PARTICIPANT_JOIN_QUIZ,
+  AuditAction.PARTICIPANT_QUIZ_PIN_SUCCESS,
+  AuditAction.PARTICIPANT_QUIZ_PIN_FAILED,
+  AuditAction.CLIENT_ERROR,
+]
+
+const ALLOWED_PUBLIC_ACTIONS_SET = new Set<string>(ALLOWED_PUBLIC_ACTIONS)
+
+const CORRELATION_ID_REGEX = /^[a-f0-9]{32}$/i
```
rschlaefli, line 20: alignment with our real correlation ID is WIP on the other branch

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
+import { AuditScope } from '@klicker-uzh/types'
+import { createHash } from 'crypto'
+import type { AuditEvent } from '../schemas/audit-event.js'
+
+export interface AuditTableEntity {
+  partitionKey: string
+  rowKey: string
+  scope: string
```
rschlaefli: where does the log come from

public (any ui) | internal (backends) | worker (hatchet/async processing)

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+  subject: string
```
rschlaefli, line 9: the user/entity that caused/triggered the event

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+  action: string
```
rschlaefli: what the action of the event was (e.g., a response was submitted)

always used via constants

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+  eventTimestamp: number
+  attributes?: string // JSON serialized
```
rschlaefli, line 12: any metadata on the event that is not in the schema already

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+  correlationId?: string
+  correlationClaims?: string // JSON serialized
+  stage?: string
+  schemaVersion: number
+  resource?: string
```
rschlaefli, line 17: the resource that is targeted (e.g., live-quiz:XY, instance:XY)

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+  partitionKey: string
+  rowKey: string
+  scope: string
+  subject: string
+  action: string
+  eventTimestamp: number
+  attributes?: string // JSON serialized
+  correlationId?: string
+  correlationClaims?: string // JSON serialized
+  stage?: string
```
rschlaefli, line 15: not really used yet, but it might be useful for e.g. when we log at multiple stages in an async workflow

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Include subset of attributes for uniqueness (avoid full object for size)
+    attributesHash: event.attributes
+      ? createHash('sha256')
+          .update(JSON.stringify(event.attributes))
+          .digest('hex')
+          .substring(0, 8)
+      : null,
+  })
+
+  return createHash('sha256').update(data).digest('hex').substring(0, 16)
```
rschlaefli, line 40: we need to observe the performance impact of this

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Generate PartitionKey for optimal distribution and querying
+ * Format: <YYYYMMDDHHmm>-<shard>
+ *
+ * This approach:
+ * - Groups events by minute for efficient range queries
+ * - Uses shard (derived from eventId) to distribute load within a time bucket
+ */
+function generatePartitionKey(timestamp: number, eventId?: string): string {
```
rschlaefli, line 83: we need to observe the performance characteristics of this

apps/audit/src/app.ts
```diff
@@ -0,0 +1,87 @@
...
+import { logger } from './utils/logger.js'
+
+const tableClient = new AuditTableClient(
+  config.AUDIT_TABLE_CONNECTION_STRING,
+  config.AUDIT_TABLE_NAME
+)
+
+const app = new Hono()
+
+const customLoggerPrintFunc = (str: string, ...rest: string[]) => {
```
rschlaefli, line 22: this connects hono (api framework) with pino (logging framework). all HTTP requests will then be logged nicely and in the same format as all of the other logs that we manually do in audit log

apps/audit/src/config.ts
```diff
@@ -0,0 +1,33 @@
...
+  APP_ORIGIN_AUTH: z.string().min(1),
+
+  // Logging configuration
+  LOG_LEVEL: z.enum(['debug', 'info', 'warn', 'error']).default('info'),
+})
+
+export type Config = z.infer<typeof configSchema>
+
+// Parse configuration once on module load to fail fast
+export const config = configSchema.parse(process.env)
```
rschlaefli, line 33: this would fail the startup of the service if we have incorrectly configured env variables. would be a good idea to do this in all services gradually

apps/frontend-pwa/src/components/liveQuiz/QuestionArea.tsx
```diff
@@ -174,6 +196,16 @@ function QuestionArea({
       correlationKey,
     })
 
+    await auditLog.log({
```
rschlaefli, line 199: first try on a UI generated audit log event

apps/frontend-pwa/src/components/liveQuiz/QuestionArea.tsx
```diff
@@ -57,6 +59,14 @@ function QuestionArea({
 }: QuestionAreaProps): React.ReactElement {
   const t = useTranslations()
 
+  const auditLog = useAuditClient({
+    assessmentMode: process.env.NEXT_PUBLIC_IS_ASSESSMENT === 'true',
```
rschlaefli, line 63: right now these two are the same, as we have no audit log outside assessment mode

apps/hatchet-worker-general/src/index.ts
```diff
@@ -107,6 +109,14 @@ async function main() {
 
   const pubSub = createPubSub({ eventTarget })
 
+  const auditClient = new AuditClient()
+
+  auditClient.log({
```
rschlaefli, line 114: these are not working so well yet and we would probably need some unique identifier per worker/similar to identify them. but I think it might still be a good idea to track the "uptime" or crashes in the backend in the audit log? just for generally critical components though..

packages/graphql/src/services/stacks.ts
```diff
@@ -2617,256 +2617,362 @@ export async function respondToQuestion(
   },
```
rschlaefli, line 2617: Did not actually change much in this file beside adding comment for future logging and moving the entire thing inside try/catch (to be able to log the error later). could undo this change if we want to keep it assessment only for now

packages/shared-components/src/hooks/useAuditClient.ts
```diff
@@ -0,0 +1,113 @@
...
+          // Wait before retry with exponential backoff (500ms, 1s, 2s)
+          if (attempt < maxRetries - 1) {
+            const delay = 500 * Math.pow(2, attempt)
+            await new Promise((resolve) => setTimeout(resolve, delay))
+          }
+        }
+      }
+
+      // All retries failed
+      setError(lastError)
```
rschlaefli, line 84: we would need to decide what we do if even though it retried it still returns an error. probably let the user know with a toast? maybe its also just their internet that is down and the main requests fail anyways

packages/util/src/audit.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const controller = new AbortController()
+    const timeoutId = setTimeout(() => controller.abort(), this.config.timeout)
+
+    try {
+      const response = await axios.request({
+        httpsAgent:
+          process.env.NODE_ENV === 'production'
+            ? undefined
+            : new https.Agent({ rejectUnauthorized: false }),
```
rschlaefli, line 128: this needs to be done for dev mode as otherwise the https validation with our self made certificate would fail in certain cases

util/wait-for-services.sh
```diff
@@ -0,0 +1,394 @@
+#!/bin/bash
```
rschlaefli, line 1: copied and extended this script from the cypress workflows -> it now waits as part of run app dependencies until the db etc. are up. could also migrate cypress workflows etc. to use this script from util/

.versionrc.js
```diff
@@ -1,46 +1,46 @@
...
-    //     filename: `deploy/charts/klicker-uzh-v2/Chart.yaml`,
-    //     updater: 'util/yaml-updater.js',
-    //   },
-    // ]
-  ),
+    // {
+    //   filename: `deploy/charts/klicker-uzh-v2/Chart.yaml`,
+    //   updater: 'util/yaml-updater.js',
+    // },
+    ...packageJsonFiles.map((filename) => ({ filename, type: 'json' })),
```
rschlaefli, line 42: much easier to parse the folders rather than having to hardcode all services -> so many were forgotten all the time and then not updated in package.json :D

_run_app_dependencies.sh
```diff
@@ -107,6 +107,12 @@ else
 	echo "Skipping pnpm run build"
 fi
 
+SERVICE_ENDPOINTS="" ./util/wait-for-services.sh || {
```
rschlaefli, line 110: empty service endpoints means dependencies only

apps/audit/test/api.test.ts
```diff
@@ -0,0 +1,284 @@
+import { describe, expect, it } from 'vitest'
```
rschlaefli, line 2: tests all work but we should have a real use case (right now we test with random data that it arrives, is processed, etc.) -> it might be interesting if we can run the assessment mode in cypress and validate the audit log output to be correct after the cypress run (maybe dump it to a file there). otherwise, the tests here are not so useful beside validating that data can be written and the api responds correctly (also why it is irrelevant what kind of data we store)

apps/audit/src/storage/entities.ts
sjschlapbach: @rschlaefli If we know that the value will be one of three strings, is there a reason, why we do not type this accordingly?

apps/audit/src/storage/entities.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Uses critical fields to ensure same event generates same ID
+ */
+function generateDeterministicEventId(event: AuditEvent): string {
+  const data = JSON.stringify({
+    scope: event.scope,
+    subject: event.subject,
+    action: event.action,
+    timestamp: event.timestamp,
+    resource: event.resource || null,
+    // Include subset of attributes for uniqueness (avoid full object for size)
```
sjschlapbach, line 31: @rschlaefli  Do I understand this correctly that we do not take a subset of the attributes (contrary to what the comment states), but actually just cut off part of the hash we generate from the attributes? Does this make sense? This seems like we go through the potentially more expensive computation to compute the hash for the large object, throw away half of what is already a compact representation, but do not care about the other fields above, which have up to 200-400 characters according to the zod validation?

apps/frontend-pwa/src/components/liveQuiz/QuestionArea.tsx
```diff
@@ -174,6 +196,16 @@ function QuestionArea({
       correlationKey,
     })
 
+    await auditLog.log({
+      action: AuditAction.PARTICIPANT_SUBMIT_RESPONSE,
+      resource: `instance:${instanceId}`,
+      scope: AuditScope.PUBLIC,
+      correlationId: correlationKey ?? undefined,
```
sjschlapbach, line 203: Maybe this is important: This `correlationKey` is not the same as the `correlationId` we use later on from within the response api through the assessment processor to track a response - the latter also contains the participant id, while this correlation key only contains the quiz id, instance id and execution ix in a format that can be validated

apps/response-api/src/index.ts
```diff
@@ -187,11 +210,19 @@ async function handleAddAssessmentResponse(
...
-    hatchetClient.events.push('create-audit-log-entry', {
-      info: `[ERROR] [AddResponse Assessment] Missing required fields in request body: ${JSON.stringify(
-        payload
-      )}`,
-    })
+    await hatchetClient.events.push<InternalAuditEvent>(
+      'create-audit-log-entry',
+      {
+        subject: `participant:unknown`,
+        correlationId: correlationKey,
```
sjschlapbach, line 218: Same note as for the frontend use - the `correlationKey` is not the same as the `correlationId` we combine from the correlation key and the participant id further down (the same holds for all the following audit log entries?)

apps/response-api/src/index.ts
```diff
@@ -275,12 +349,20 @@ async function handleAddAssessmentResponse(
...
-    info: `[AddResponse Assessment] Response-API received response for instance ${instanceId} in live quiz ${liveQuizId} from participant ${user.sub}: ${JSON.stringify(
-      response
-    )}`,
-  })
+  await hatchetClient.events.push<InternalAuditEvent>(
+    'create-audit-log-entry',
+    {
+      scope: AuditScope.INTERNAL,
+      subject: `participant:${user.sub}`,
+      correlationId: correlationKey,
```
sjschlapbach, line 366: (same as above) - from here onwards, the correlationId is available and should be used

PR #4893, opened 2025-09-10
feat: add new processor for student assessment responses

sjschlapbach:
## Summary by CodeRabbit

* **New Features**
  * Assessment mode: JWT-verified participants, submission correlation IDs, vote deduplication, and Redis-backed duplicate checks.
  * Background processing: durable tasks to process and aggregate assessment responses; per-participant live-quiz responses persisted with timing, scoring, and leaderboard updates.

* **Improvements**
  * Centralized scoring with separate correctness/bonus, optional rounding, and broadened question-type handling.
  * Live-quiz lifecycle: block start/close timestamps and richer per-instance caching.
  * Frontend: response payloads include correlationId and assessment endpoint updates.

* **Chores**
  * Env and dependency updates; packaging/runtime copy adjustments.

### DIFF
--- a/apps/frontend-pwa/.env.assessment
+++ b/apps/frontend-pwa/.env.assessment
@@ -3,8 +3,7 @@ API_DOMAIN=assessment-api$COOKIE_DOMAIN
 
 NEXT_PUBLIC_AVATAR_BASE_PATH="https://sos-ch-dk-2.exo.io/klicker-prod/avatars"
 NEXT_PUBLIC_IMAGE_BASE_PATH="https://sos-ch-dk-2.exo.io/klicker-prod/img"
-NEXT_PUBLIC_ADD_RESPONSE_URL="https://backend-responses$COOKIE_DOMAIN/api/AddResponse"
-NEXT_PUBLIC_ADD_RESPONSE_V2_URL="https://response-api$COOKIE_DOMAIN/AddResponse"
+NEXT_PUBLIC_ADD_RESPONSE_URL="https://assessment-response-api$COOKIE_DOMAIN/AddResponse"
 NEXT_PUBLIC_API_URL="https://$API_DOMAIN/api/graphql"
 NEXT_PUBLIC_API_URL_SSR="http://klicker-v2-prod-klicker-uzh-v2-backend-graphql:3000/api/graphql"
 NEXT_PUBLIC_WEB_PUSH_PUBLIC_KEY="BPz0NNgACtulol5iYwmnFzzL-gkTUQ2jZrXsdrr9oEfhZUByWw8n_lu6tiDGmrbiJqRT-CYw0p9rextvtknuQAE"
--- a/apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
+++ b/apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
@@ -41,10 +41,11 @@ export async function processAssessmentResponse(
   },
   ctx: Context<JsonObject, {}> | DurableContext<JsonObject, {}>
 ) {
-  ctx.logger.info(
-    'ProcessAssessmentResponse function processing a message',
-    message
-  )
+  ctx.logger.info('ProcessAssessmentResponse: received message', {
+    correlationId: message.correlationId,
+    sessionId: message.sessionId,
+    instanceId: message.instanceId,
+  })
   ctx.v1.events.push('create-audit-log-entry', {
     correlationId: message.correlationId,
     info: `[AddResponse Assessment] Processing response for instance ${message.instanceId} by participant ${message.participantId}.`,
@@ -78,7 +79,14 @@ export async function processAssessmentResponse(
   const response = message.response
 
   if (!response) {
-    ctx.logger.error(`Missing response: ${JSON.stringify(message)}`)
+    ctx.logger.error(
+      'Missing response ' +
+        JSON.stringify({
+          correlationId: message.correlationId,
+          sessionId: message.sessionId,
+          instanceId: message.instanceId,
+        })
+    )
     throw new Error('Missing response')
   }
 
@@ -88,9 +96,10 @@ export async function processAssessmentResponse(
 
   // if the instance info is not available, return that the corresponding cache data is not available
   if (!instanceInfo || Object.keys(instanceInfo).length === 0) {
-    ctx.logger.info(
-      `Element instance metadata for instance ${message.instanceId} not found.`
-    )
+    ctx.logger.info('Element instance metadata not found', {
+      correlationId: message.correlationId,
+      instanceId: message.instanceId,
+    })
     ctx.v1.events.push('create-audit-log-entry', {
       correlationId: message.correlationId,
       info: `[AddResponse Assessment] Instance metadata for instance ${message.instanceId} not found.`,
@@ -113,7 +122,10 @@ export async function processAssessmentResponse(
   } = instanceInfo
 
   if (blockClosedAt && Number(responseTimestamp) > Number(blockClosedAt)) {
-    ctx.logger.info('Response received after element block was closed')
+    ctx.logger.info('Response received after element block was closed', {
+      correlationId: message.correlationId,
+      instanceId: message.instanceId,
+    })
     ctx.v1.events.push('create-audit-log-entry', {
       correlationId: message.correlationId,
       info: `[AddResponse Assessment] Response received after block of element instance ${message.instanceId} was closed at ${new Date(blockClosedAt)}.`,
@@ -129,9 +141,7 @@ export async function processAssessmentResponse(
       parsedSolutions = JSON.parse(solutions)
     }
   } catch (e) {
-    ctx.logger.error(
-      `Error parsing solutions (Error: ${JSON.stringify(e)}, Message: ${JSON.stringify(message)})`
-    )
+    ctx.logger.error(`Error parsing solutions: ${String(e)}`)
     throw new Error('Error parsing solutions')
   }
 
@@ -150,7 +160,14 @@ export async function processAssessmentResponse(
     case ElementType.KPRIM: {
       // if response choices are not defined, return early
       if (!response.choices) {
-        ctx.logger.error(`Missing response choices: ${JSON.stringify(message)}`)
+        ctx.logger.error(
+          'Missing response choices ' +
+            JSON.stringify({
+              correlationId: message.correlationId,
+              sessionId: message.sessionId,
+              instanceId: message.instanceId,
+            })
+        )
         ctx.v1.events.push('create-audit-log-entry', {
           correlationId: message.correlationId,
           info: `[AddResponse Assessment] Response to choices question (instance id ${message.instanceId}) does not contain choices.`,
@@ -181,7 +198,14 @@ export async function processAssessmentResponse(
     case ElementType.NUMERICAL: {
       // if response value is not defined, return early
       if (typeof response.value === 'undefined' || response.value === null) {
-        ctx.logger.error(`Missing response value: ${JSON.stringify(message)}`)
+        ctx.logger.error(
+          'Missing response value ' +
+            JSON.stringify({
+              correlationId: message.correlationId,
+              sessionId: message.sessionId,
+              instanceId: message.instanceId,
+            })
+        )
         ctx.v1.events.push('create-audit-log-entry', {
           correlationId: message.correlationId,
           info: `[AddResponse Assessment] Response to numerical question (instance id ${message.instanceId}) does not contain value.`,
@@ -210,7 +234,14 @@ export async function processAssessmentResponse(
     case ElementType.FREE_TEXT: {
       // if response value is not defined, return early
       if (typeof response.value !== 'string') {
-        ctx.logger.error(`Missing response value: ${JSON.stringify(message)}`)
+        ctx.logger.error(
+          'Missing response value ' +
+            JSON.stringify({
+              correlationId: message.correlationId,
+              sessionId: message.sessionId,
+              instanceId: message.instanceId,
+            })
+        )
         ctx.v1.events.push('create-audit-log-entry', {
           correlationId: message.correlationId,
           info: `[AddResponse Assessment] Response to free text question (instance id ${message.instanceId}) does not contain value.`,
@@ -239,7 +270,12 @@ export async function processAssessmentResponse(
       // if response selection is not defined, return early
       if (!response.selection) {
         ctx.logger.error(
-          `Missing response selection: ${JSON.stringify(message)}`
+          'Missing response selection ' +
+            JSON.stringify({
+              correlationId: message.correlationId,
+              sessionId: message.sessionId,
+              instanceId: message.instanceId,
+            })
         )
         ctx.v1.events.push('create-audit-log-entry', {
           correlationId: message.correlationId,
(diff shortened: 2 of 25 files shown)


packages/grading/src/index.ts
```diff
@@ -272,62 +272,113 @@
...
+  // add the points awarded for correct answers
+  awardedPoints += correctnessPoints + bonusPoints
+
+  // depending on the base points setting, compute the final awarded points
   awardedPoints += basePoints ? (defaultPoints ?? 0) : 0
 
-  return Math.round(awardedPoints)
+  // if desired, round the result to the nearest integer
+  return roundedResult ? Math.round(awardedPoints) : awardedPoints
 }
```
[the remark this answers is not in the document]
sjschlapbach: Multipliers are only applied to the correctness and bonus points in live quizzes, base points are not affected by the multiplier by design

packages/prisma/src/prisma/schema/migrations/20250910170505_live_quiz_response/migration.sql
```diff
@@ -0,0 +1,34 @@
...
+-- CreateTable
+CREATE TABLE "public"."LiveQuizResponse" (
+    "id" SERIAL NOT NULL,
+    "submittedAt" TIMESTAMP(3) NOT NULL,
+    "response" JSONB NOT NULL,
+    "timeSpent" REAL NOT NULL,
+    "correctness" "public"."ResponseCorrectness" NOT NULL,
+    "basePoints" FLOAT(2) NOT NULL,
+    "correctnessPoints" FLOAT(2) NOT NULL,
+    "bonusPoints" FLOAT(2) NOT NULL,
```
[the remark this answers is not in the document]
sjschlapbach: Rounding drift is not an issue here, since these values are never update, but only stored once. The choice for FLOAT was conscious here, since a large number of rows is expected to be added to this table and using the numeric or decimal types instead would mean a significant overhead in storage space that is not needed.

apps/frontend-pwa/.env.assessment
```diff
@@ -3,8 +3,7 @@ API_DOMAIN=assessment-api$COOKIE_DOMAIN
 
 NEXT_PUBLIC_AVATAR_BASE_PATH="https://sos-ch-dk-2.exo.io/klicker-prod/avatars"
 NEXT_PUBLIC_IMAGE_BASE_PATH="https://sos-ch-dk-2.exo.io/klicker-prod/img"
-NEXT_PUBLIC_ADD_RESPONSE_URL="https://backend-responses$COOKIE_DOMAIN/api/AddResponse"
-NEXT_PUBLIC_ADD_RESPONSE_V2_URL="https://response-api$COOKIE_DOMAIN/AddResponse"
+NEXT_PUBLIC_ADD_RESPONSE_URL="https://assessment-response-api$COOKIE_DOMAIN/AddResponse"
```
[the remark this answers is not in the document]
sjschlapbach: During development two different commands are used to start the application either in assessment mode (with the corresponding frontend application at assessment.klicker.com and the associated response-api at assessment-repsonse-api.com. During deployment, the same code will be deployed on two different pods, where we use the assessment flag once (or multiple times) to create the assessment frontend and response api containers and the standard ones separately.

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+        submittedAt: new Date(responseTimestamp),
+        response,
+        timeSpent: -1, // TODO: set this in future improvements
+        correctness:
+          computedCorrectness === null || computedCorrectness === 1
+            ? ResponseCorrectness.CORRECT
+            : computedCorrectness === 0
+              ? ResponseCorrectness.WRONG
+              : ResponseCorrectness.PARTIAL,
+        basePoints: awardedBasePoints,
```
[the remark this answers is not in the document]
sjschlapbach: Questions without sample solution return null as a grading decision. Since we want to have the corresponding fields in the live quiz response as required for the moment, we decided to store responses to these questions with correct and no points (or only base points - depending on the setting)

packages/hatchet-tasks/src/index.ts
```diff
@@ -1,18 +1,18 @@
-import { HatchetClient } from '@hatchet-dev/typescript-sdk'
+import { HatchetClient, Priority } from '@hatchet-dev/typescript-sdk'
 
 export type SimpleInput = {}
 
 export function prepareHatchetTasks(hatchet: HatchetClient) {
-  // TODO: implement
-  const doSomethingTask = hatchet.task({
-    name: 'do-something',
+  const createAuditLogEntryTask = hatchet.durableTask({
```
rschlaefli: does it have to be a durable task? I don't grasp the full difference between non-durable and durable yet but I think tasks are also persisted/retried even if they are not durable and durable is only if we need to use/update task context etc

packages/hatchet-tasks/src/index.ts
```diff
@@ -1,18 +1,18 @@
...
+    name: 'create-audit-log-entry',
     retries: 3,
-    fn: async (input: SimpleInput) => {
-      return {
-        hello: 'world',
-      }
+    defaultPriority: Priority.MEDIUM,
```
rschlaefli: we could maybe even make it low priority (if prio is only for scheduling, we would rather process responses than logging and do the logging afterwards, as long as it is logged at some point?

apps/response-api/src/index.ts
```diff
@@ -144,18 +115,123 @@ async function handleAddResponse(req: IncomingMessage, res: ServerResponse) {
...
+    `lq:${sessionId}:i:${instanceId}:votes`,
+    correlationId
+  )
+  if (votes) {
+    console.log(
+      `Participant with correlationId ${correlationId} already answered instance ${instanceId} in session ${sessionId}`
+    )
+    hatchet.events.push('create-audit-log-entry', {
+      correlationId,
+      info: `[AddResponse Assessment] Participant with correlationId ${correlationId} already tried to answer instance ${instanceId} in session ${sessionId} again.`,
```
rschlaefli:
```suggestion
      info: `[AddResponse Assessment] Participant with correlationId ${correlationId} tried to answer instance ${instanceId} in session ${sessionId} again.`,
```

apps/response-api/src/index.ts
```diff
@@ -144,18 +115,123 @@ async function handleAddResponse(req: IncomingMessage, res: ServerResponse) {
...
+    })
+  }
+
+  const message = {
+    correlationId,
+    participantId: user.sub,
+    sessionId: String(sessionId),
+    instanceId: String(instanceId),
+    response, // pass through as-is; worker validates
+    responseTimestamp: Date.now(),
```
rschlaefli: maybe we could even create the date on the first line of the function -> we would make sure any code computation time etc. is also not included

apps/response-api/src/index.ts
```diff
@@ -144,18 +115,123 @@ async function handleAddResponse(req: IncomingMessage, res: ServerResponse) {
...
+  }
+
+  // start the processing of an assessment response
+  console.log(
+    `Pushing event ${'response-received:assessment'} with payload`,
+    message
+  )
+
+  await hatchet.events.push('response-received:assessment', message)
```
rschlaefli: not sure how it works with hatchet, but what if this fails? do we need to: try again/catch the error/log into the audit log that it failed? if it fails then the audit log would probably also fail, then we might need to fall back to sending to the audit log via network (might make sense actually to have the audit log call in a function, send the event, if sending fails, automatically do it via network request)

or do we want to log this in general to the audit log as "response added to queue"? maybe, maybe not, might be too many audit log events at some point

packages/grading/src/index.ts
```diff
@@ -272,62 +272,113 @@ export function gradeQuestionCaseStudy({
...
+  responseTimestamp: number
+  maxBonus: number
+  timeToZeroBonus?: number
+  getsMaxPoints?: boolean
+  defaultPoints?: number
+  defaultCorrectPoints?: number
+  pointsPercentage?: number | null
+  basePoints: boolean
+  pointsMultiplier?: number | string | null
+  roundedResult: boolean
```
rschlaefli, line 344: in which cases would we use rounded and not rounded? are there specific use cases or will we only use one/this parameter is there for future usage?

probably we would not use rounded as the decimals are limited in the database, and then maybe just round in the ui? or would you do it differently?

apps/hatchet-worker-response-processor/src/index.ts
```diff
@@ -31,11 +38,33 @@ export const processAuthenticatedResponseTask = hatchet.durableTask({
...
+  onEvents: ['response-received:assessment'],
+  fn: processAssessmentResponse,
+})
+
+export const aggregateAssessmentResponsesTask = hatchet.durableTask({
+  name: 'aggregate-assessment-responses',
+  retries: 1,
+  defaultPriority: Priority.MEDIUM,
+  concurrency: {
+    expression: 'input.instanceId', // use the instance id as a concurrency key to ensure only a single aggregation task is running per instance
```
rschlaefli, line 82: this is automatically extracted from the input to the function based on the "path" we write here as a string? interesting

apps/hatchet-worker-response-processor/src/helpers.ts
```diff
@@ -0,0 +1,455 @@
+import {
```
rschlaefli, line 1: someday could probably refactor this a bit to reduce duplication

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+  })
+  ctx.v1.events.push('create-audit-log-entry', {
+    correlationId: message.correlationId,
+    info: `[AddResponse Assessment] Processing response for instance ${message.instanceId} by participant ${message.participantId}.`,
+  })
+
+  try {
+    assert(!!redisExec)
+  } catch (e) {
+    ctx.logger.error(`Redis connection error: ${JSON.stringify(e)}`)
```
rschlaefli, line 55: do we need to audit log in these failure cases (might be that redis is down shortly but back up for another user)? the cleaner alternative might be that we have a hatchet task defined that is triggered after the failure of this task or maybe after all retries failed. otherwise we have to put so many audit log statements?

https://docs.hatchet.run/home/on-failure-tasks

what do you think?

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+  if (!response && instanceInfo.type !== ElementType.CONTENT) {
+    ctx.logger.error(
+      'Missing response ' +
+        JSON.stringify({
+          correlationId: message.correlationId,
+          sessionId: message.sessionId,
+          instanceId: message.instanceId,
+        })
+    )
+    throw new NonRetryableError('Missing response')
```
rschlaefli, line 91: audit log as invalid response received&will not be retried?

rschlaefli: not sure, concerning the comment above, if this would also automatically trigger the on failure case and thus send it to the audit log automatically. would be nice..

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+    choiceCount,
+    basePoints,
+    defaultPoints,
+    pointsMultiplier,
+    blockExecution,
+    blockClosedAt,
+  } = instanceInfo
+
+  if (blockClosedAt && Number(responseTimestamp) > Number(blockClosedAt)) {
+    ctx.logger.info('Response received after element block was closed', {
```
rschlaefli: if we make a function/helper for audit logging, we could have the logic in there that it always does logging to context plus sending an event? would reduce having the amount of times we do both

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+  // extract the relevant information from the redis cache
+  const liveQuizKey = `lq:${message.sessionId}`
+  const instanceKey = `${liveQuizKey}:i:${message.instanceId}`
+  const responseTimestamp = message.responseTimestamp
+  const response = message.response
+
+  // get live quiz and instance information from redis cache
+  const instanceInfo = await redisExec.hgetall(`${instanceKey}:info`)
+
+  if (!response && instanceInfo.type !== ElementType.CONTENT) {
```
rschlaefli, line 82: in the case of content, do we need to process any further or can we end here already? this should not even be possible anyways, right? so it could maybe be ||, if no response, or content (or flashcard)? or would you allow adding them including tracking responses to them?

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+  // ! Step 2: Switch between different types, validate response and compute awarded points and XP
+
+  let parsedSolutions = undefined
+  try {
+    if (solutions) {
+      parsedSolutions = JSON.parse(solutions)
+    }
+  } catch (e) {
+    ctx.logger.error(`Error parsing solutions: ${String(e)}`)
+    throw new Error('Error parsing solutions')
```
rschlaefli: here is probably non retryable, if the parsing fails, trying again will not solve it. but then we should make sure that it is logged to audit as an invalid solution including the full solutions field.

rschlaefli: or we could log to audit that solutions were invalid, but continue processing as if there were no solutions (e.g., still giving base points)? that would be better for the students, as the quiz can still run, but it would be intransparent

rschlaefli: nevermind, it could be retryable, in case something in redis changes before the retry anyways. but would hold for any parsing based on redis data also later in the code

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+          pointsMultiplier,
+          parsedSolutions,
+        })
+      computedCorrectness = pointsPercentage
+      awardedCorrectnessPoints = correctnessPoints
+      awardedBonusPoints = bonusPoints
+      awardedXp = xpAwarded
+
+      break
+    }
```
rschlaefli, line 275: should we have a default case that fails and logs with unsupported question type? not sure when this could ever happen, mostly due to future dev error

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
```diff
@@ -0,0 +1,578 @@
...
+    })
+  } catch (e) {
+    ctx.logger.error(
+      `Error during live quiz response creation: ${JSON.stringify(e)}`
+    )
+    ctx.v1.events.push('create-audit-log-entry', {
+      correlationId: message.correlationId,
+      info: `[AddResponse Assessment] Failed to create live quiz response for instance ${message.instanceId} and participant ${message.participantId}.`,
+    })
+    throw new NonRetryableError(
```
rschlaefli: why is this non retryable in general? might be a database outage for a quick period. I think here we need a catch for specific types of errors, as the "unique key etc." would be non retryable, but the "network error" would and needs to be retried.

sjschlapbach: true, updated this to a regular error

rschlaefli (approved):
Verified locally

apps/response-api/src/index.ts
sjschlapbach: add TODO that we send an audit log entry via network request (or hatchet)

apps/hatchet-worker-response-processor/src/assessmentProcessor.ts
sjschlapbach: check if response is sent for content elements, try to find consistent solution for all element types

apps/hatchet-worker-response-processor/src/index.ts
sjschlapbach: hopefully this works, very hard to test locally, is noted to be verified during load tests

PR #4899, opened 2025-09-12
enhance: correctly display errors from response-api in student frontend and resolve race conditions

sjschlapbach:
## Summary by CodeRabbit

- New Features
  - Clear, status-based toasts for submission outcomes and a “Responded at” timestamp in the header.
  - Autosave of in-progress answers with safer, race‑safe restoration.

- Bug Fixes
  - More reliable submission flow with per-type validation; prevents saving invalid answers and avoids stale UI state.
  - Duplicate submissions now detected and reported to users.

- Chores
  - Tightened allowed origins to HTTPS and added the assessment domain.

### DIFF
--- a/packages/util/src/jwt.ts
+++ b/packages/util/src/jwt.ts
@@ -22,12 +22,13 @@ export async function signJWT(
   options: {
     algorithm?: 'HS256'
     expiresIn?: string | number
-  } = {}
+  } = {},
+  issuedAt?: Date
 ): Promise<string> {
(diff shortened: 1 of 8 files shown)


packages/util/src/jwt.ts
```diff
@@ -22,12 +22,13 @@ export async function signJWT(
   options: {
     algorithm?: 'HS256'
     expiresIn?: string | number
```
rschlaefli, line 24: it would be cleaner to add it here so it is a named property in the object

PR #4915, opened 2025-09-15
fix(packages/graphql): correctly skip invalid submissions for case study question in assessment live quiz response aggregation

sjschlapbach:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed case-study response aggregation to avoid errors and ensure accurate counts.
  * Late responses received after a block closes are now skipped without causing failures.
  * Duplicate submissions are detected and ignored to prevent duplicate votes and inconsistent results.

* **Chores**
  * Added a maintenance script to recompute and persist per-element quiz results after a block closes for reliable historical data.


apps/hatchet-worker-response-processor/src/processors/assessmentProcessor.ts
```diff
@@ -125,9 +125,11 @@ export async function processAssessmentResponse(
...
 
   if (blockClosedAt && Number(responseTimestamp) > Number(blockClosedAt)) {
-    throw new NonRetryableError(
-      `Response received after block of element instance ${message.instanceId} was closed at ${new Date(blockClosedAt)}.`
+    ctx.logger.error(
+      `[CANCEL] [AddResponse Assessment] Response received after block of element instance ${message.instanceId} was closed at ${new Date(Number(blockClosedAt))}.`
     )
+    ctx.cancel()
+    return
   }
```
sjschlapbach, line 133: Cancel the task (-> success status but cancellation in timeline) if the response arrived after the block was closed

apps/hatchet-worker-response-processor/src/processors/assessmentProcessor.ts
```diff
@@ -371,6 +373,25 @@ export async function processAssessmentResponse(
...
+    },
+  })
+
+  if (existingVote) {
+    ctx.logger.error(
+      `[CANCEL] [AddResponse Assessment] Participant ${message.participantId} has already submitted a response for instance ${message.instanceId} and block execution ${blockExecution}.`
+    )
+    ctx.cancel()
+    return
+  }
```
sjschlapbach, line 393: We first check for an existing live quiz response to determine if we should create a new one instead of only handling arbitrary failures of the creation function (-> success status, but cancellation in timeline)

packages/graphql/src/services/liveQuizzes.ts
```diff
@@ -1553,7 +1553,7 @@ function aggregateLiveQuizResponses({
                         criterionId
                       ] === 'undefined'
                     ) {
-                      return
+                      return acc
```
sjschlapbach, line 1558: by returning early we do not add invalid data to the results -> student submissions might be processed partially, however

PR #4927, opened 2025-09-18
enhance(packages/graphql): instead of unlinking, set expiration date on instance keys within block closure operation

sjschlapbach:
## Summary by CodeRabbit

- Chores
  - Improved cache handling for live quizzes by transitioning to time-based expiration, enhancing stability and reducing potential disruptions.
  - Background reliability improvement with no changes to the user interface or workflows.


sjschlapbach:
@rschlaefli The function has been verified locally and works correctly - the cache keys are now not removed anymore when the block is closed and the results are aggregated 5 min after block closure click, but their expiry date is set to 24h instead -> is automatically removed after that time

PR #4981, opened 2025-12-12
chore(packages/graphql): add script to transfer activities between courses

jabbadizzleCode:
## Summary by CodeRabbit

* **New Features**
  * Added ability to transfer activities (live quizzes, practice quizzes, micro‑learnings, group activities, and chatbots) between courses with ownership validation.
  * Transfers run atomically with a timeout, log progress, emit cache invalidation events, and automatically recompute permissions for moved items; clear errors are returned if courses are missing or owners differ.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


packages/graphql/src/scripts/transferActivitiesBetweenCourses.ts
```diff
@@ -0,0 +1,233 @@
...
+      practiceQuizzes: true,
+      microLearnings: true,
+      groupActivities: true,
+      chatbots: true,
+    },
+  })
+  const newCourse = await prisma.course.findUnique({
+    where: { id: newCourseId },
+    include: {
+      owner: true,
```
sjschlapbach: detail: since we only compare the ownerIds, there is no need to fetch the owner for the old and new courses, since the ownerId is should already be available on the course table entry itself as the join field :)

packages/graphql/src/scripts/transferActivitiesBetweenCourses.ts
```diff
@@ -0,0 +1,233 @@
...
+    throw new Error(`Destination course with ID ${newCourseId} not found`)
+  }
+
+  if (oldCourse.owner.id !== newCourse.owner.id) {
+    throw new Error(
+      `Old course owner and new course owner are not the same user (ID: ${oldCourse.ownerId})`
+    )
+  }
+  const oldUserId = oldCourse.owner.id
+  const newUserId = newCourse.owner.id
```
sjschlapbach: detail: after removing the corresponding ids from the permissions updates, we can also remove these variables

packages/graphql/src/scripts/transferActivitiesBetweenCourses.ts
```diff
@@ -0,0 +1,233 @@
...
+        emitter.emit('invalidate', {
+          typename: 'LiveQuiz',
+          id: liveQuizId,
+        })
+
+        // recompute derived permissions for the old user (-> remove owner permissions)
+        await recomputeDerivedPermissions(
+          {
+            liveQuizId,
+            userId: oldUserId,
```
sjschlapbach: This would result in issues, if there are any users that have access to the course and thereby indirect access to the activity contained in the course. Correspondingly, we would have to combine this call and the next one to the recomputeDerivedPermissions function to a single call with only the liveQuizId being defined. In this way, derived permissions on the activity that are no longer valid (and new derived permissions that need to be created through the new course assignment) should be created automatically.

sjschlapbach: The same needs to be applied to all activities below as well

packages/graphql/src/scripts/transferActivitiesBetweenCourses.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  if (oldCourse.owner.id !== newCourse.owner.id) {
+    throw new Error(
+      `Old course owner and new course owner are not the same user (ID: ${oldCourse.ownerId})`
+    )
+  }
+  const oldUserId = oldCourse.owner.id
+  const newUserId = newCourse.owner.id
+
+  await prisma.$transaction(
```
sjschlapbach, line 46: comment only: If the runtime of this transaction turns out to be a problem, I think we could also think about only wrapping the ownerId updates of the corresponding activities and their corresponding permission updates in a transaction, right?

sjschlapbach (changes requested):
The PR already looks good, some minor fixes are still required to ensure that the derived permissions remain correct. Additionally, the coderabbit comments are also smart in my opinion (e.g. the typing of the ids as a set) to ensure that no id is defined twice (might result in an error during the update?).

PR #5036, opened 2026-03-19
enhance(apps/chat): add litellm proxy

TheHummel:
- Adds Litellm as a proxy service
- Includes patch for ai sdk fetch function for createOpenai for Responses API which add missing fields (type: 'message', status: 'completed') for restrictive providers

## Summary by CodeRabbit

* **New Features**
  * Per-chatbot custom OpenAI API key and endpoint support.
  * Local model proxy service added for centralized model routing.

* **Improvements**
  * More reliable model selection and streaming; startup warnings when key/env vars are missing.
  * Chat model registry now accepts models without an explicit API version.

* **Infrastructure**
  * Proxy service added with health checks; environment tracking extended to include upstream proxy variables.


apps/chat/src/app/api/chatbots/[chatbotId]/chat/route.ts
```diff
@@ -33,44 +33,65 @@ const isDevLogging = process.env.NODE_ENV === 'development'
...
-    model,
-    debug: {
-      baseUrl: baseUrl,
-      hasCustomEndpoint,
-      hasCustomApiKey,
-    },
-  }
+  // Default: route through LiteLLM
+  return createOpenAI({
+    baseURL: process.env.LITELLM_BASE_URL,
```
Comment (rschlaefli) [inline apps/chat/src/app/api/chatbots/[chatbotId]/chat/route.ts]: do not hardcode the litellm naming -> use jsut OPENAI_BASE_URL and OPENAI_API_KEY -> this makes it compatible with any openai endpoint

rschlaefli (changes requested):

PR #5045, opened 2026-03-26
feat(chat): add image attachment support

TheHummel:
## Summary
- add end-to-end image attachment support to chat, including composer upload flow, attachment validation, and support for multiple images per message
- add AI-generated image descriptions for prior-turn model context, model capability/config updates, and the related credit-accounting and prompt/config wiring
- persist attachments with stable ordering, preview-safe history payloads, lazy full-image hydration, edit/regenerate preservation, and a preview tile plus on-demand viewer UX for persisted history attachments
- generate persisted preview bytes server-side with `sharp` so the `/chat` request only carries full image bytes and avoids the local dev parse failure we hit during verification

## Verification
- `pnpm --filter @klicker-uzh/chat check`
- `pnpm --filter @klicker-uzh/chat test:run`
- Browser verified on the seeded chatbot flow: login, attach a real PNG plus text, send successfully, reload the thread, and open the persisted history image in the viewer

## Notes
- includes Prisma schema/migration work for attachment ordering and preview persistence
- includes the updated implementation plan in `project/CHAT_IMAGE_IMPROVEMENTS.md`


packages/prisma/src/prisma/schema/migrations/20260403075358_add_chat_attachments/migration.sql
```diff
@@ -0,0 +1,21 @@
+-- CreateEnum
+CREATE TYPE "public"."ChatAttachmentType" AS ENUM ('IMAGE');
+
+-- CreateTable
+CREATE TABLE "public"."ChatAttachment" (
+    "id" UUID NOT NULL,
+    "type" "public"."ChatAttachmentType" NOT NULL,
+    "imageBase64" TEXT,
```
jabbadizzleCode, line 8: Instead of storing the entire image as a base64 string in PostgreSQL, maybe consider to store the images on an Azure Blob Storage and store only the URL in PostgreSQL (cheaper) 😄.

jabbadizzleCode (approved):
Looks good to me :). Before merging, maybe consider using Azure Blob Storage for the images instead of PostgreSQL (see comment).

PR #5087, opened 2026-05-07
chore(apps/chat): update chatbot styling

TheHummel:
## Summary by CodeRabbit

* **New Features**
  * Threads grouped by date sections (Today, Yesterday, This Week, Earlier)

* **Improvements**
  * Assistant reasoning concatenates and formats sections with clearer blank-line separation (including aborted responses)
  * Reasoning renders with improved spacing and Markdown formatting
  * Image descriptions render with normalized Markdown for clearer display
  * Composer input cursor/interaction and thread welcome styling refined; edit controls simplified (direct icon buttons)
  * Thread timestamps now update accurately
  * Markdown rendering supports single-dollar math mode; math-tag normalization helper made reusable

* **Tests**
  * Thread fixtures updated to include timestamps


apps/chat/src/components/thread-image-viewer-modal.tsx
```diff
@@ -1,7 +1,7 @@
 'use client'
 
-import { Markdown } from '@klicker-uzh/markdown'
```
rschlaefli, line 3: @TheHummel it does not make sense to NOT use our shared markdown package -> with this package we make sure that it looks the same way in any place in the app where we render markdown. if you do need other things that this markdown package does not offer, you would have to propose changes to the package. I would like to prevent duplicating markdown rendering and reinstalling packages across the app.

apps/chat/src/components/thread-list.tsx
```diff
@@ -25,26 +24,94 @@ const ThreadListItems: FC = () => {
   const router = useRouter()
   const { threads, deleteThread } = useChatStore()
 
+  const groupedThreads = groupThreadsByDate(threads)
```
rschlaefli: it looks to me like this computation could be smart to do as a custom hook that can be memoized rather than just calling a function here (though future react compiler might handle that for us)

apps/chat/src/components/markdown-text.tsx
```diff
@@ -12,16 +12,20 @@ import {
 } from '@assistant-ui/react-markdown'
```
rschlaefli, line 12: this custom component should be removed with regards to our markdown package unless there is such a large difference between the two

apps/chat/package.json
```diff
@@ -41,6 +41,7 @@
     "pg": "8.16.3",
     "react": "19.1.2",
     "react-dom": "19.1.2",
+    "react-markdown": "10.1.0",
```
rschlaefli: this will not be needed if we use the markdown package

rschlaefli (changes requested):

apps/chat/src/components/thread-image-viewer-modal.tsx
TheHummel: Yes, that makes sense,  the reason was the Markdown wasn't working for single $ signs but will propose change to shared markdown package instead

apps/chat/src/components/markdown-text.tsx
TheHummel: i think the assistant-ui/react-markdown imports are needed for rendering while streaming and i think they ensure not all the elements are re-rendered with every token i.e. its a assistant-ui import to ensure the streaming renders efficiently

apps/chat/src/components/thread-list.tsx
TheHummel: I agree, memoization makes sense here

PR #5173, opened 2026-07-15
fix(assessment): harden verifiable performance reports

jabbadizzleCode:
## Summary

This is a stacked remediation PR for #5141 and targets `export-assessment-performance-insights`.

It hardens assessment-report issuance, verification, privacy, revocation, and frontend behavior. Reports use server-computed assessment data and identify students through their accepted course invitation. An Edu-ID login or UZH email address is not required.

### Credential lifecycle

- Stores immutable, versioned assessment snapshots with canonical SHA-256 hashes.
- Models `ACTIVE`, `REVOKED`, and `SUPERSEDED` lifecycle states.
- Enforces one active report per participant, course, and credential type.
- Uses serializable transactions, bounded retries, and idempotent issuance.
- Prevents unchanged revoked reports from being reissued.
- Rewrites the original, undeployed credential migration instead of adding a follow-up migration.

### Identity and issuance

- Requires an active participant, active participation, assessment-enabled course, and accepted course invitation.
- Uses the earliest valid accepted invitation email, including non-UZH addresses.
- Does not fall back to participant profile data, usernames, or client-supplied claims.
- Keeps issuance participant-only without requiring an Edu-ID-specific scope.
- Computes all report results and comparison data on the server.

### Verification and privacy

- Exposes a minimal public verification projection.
- Redacts claims for revoked, superseded, malformed, or unsupported snapshots.
- Uses fragment-based verification URLs so tokens are not included in page requests.
- Sends public verification tokens and lecturer email searches through POST requests.
- Requires course `ADMIN` permission for report records and full account access plus `ADMIN` permission for revocation.
- Adds searchable, filterable, and paginated lecturer records with a count query.

### Frontends

- Generates standalone HTML reports exclusively from the issued server snapshot.
- Escapes all dynamic HTML, awaits QR generation, and embeds the official UZH logo.
- Adds explicit PWA actions to view, download, and refresh an issued report.
- Adds localized verification states in English and German.
- Completes the lecturer workflow with search, status filters, pagination, truncated tokens, revocation confirmation, and visible errors.
- Removes the obsolete token-in-path verification route.

### Development fixture

- Adds a development-only Assessment Course with 30 participants.
- Includes one completed assessment live quiz with differing participant results.
- Includes accepted invitation emails, including non-UZH addresses.
- Supports manual testing of assessment results, point corrections, reports, and verification.
- Leaves Cypress and Playwright test-environment seeding unchanged.

## Verification

- `pnpm run check:all` passed.
- `pnpm run build` passed through the pre-push hook.
- GraphQL codegen ran twice and left a clean tree.
- Assessment snapshot tests passed: 7/7.
- Development assessment seed executed successfully and remained idempotent.
- Student login, report issuance, report viewing, and HTML download were verified in the browser.
- The downloaded report resolved to an active public verification record.
- English and German verification pages were verified, including a mobile viewport.
- Lecturer report count, listing, email search, and revocation confirmation were verified.
- An independent strict review found no actionable P0-P2 blockers.

## Review Notes

- The PR must target `export-assessment-performance-insights`, not `v3`.
- The credential migration is assumed not to have been deployed.
- Accepted course invitations are the identity authority; no `@uzh.ch` restriction is applied.
- Existing Edu-ID requirements for assessment sessions and response submission remain unchanged.
- Hosted CI must pass before either PR is merged.


rschlaefli:
Follow-up commit 623eab1a7 addresses the three Greptile P2 findings confirmed by an independent review: revocation existence-oracle (auth check now precedes existence disclosure; both cases return NOT_FOUND), bounded P2034 retry around the getEquivalentActiveRecord fallback, and score clamping in the histogram bin lookup (all four call sites).

PR #5322, opened 2026-08-06
feat(feature-flags): add shared GrowthBook foundation

jabbadizzleCode:
## Description

Introduces the shared, typed GrowthBook foundation for Klicker frontend and
backend consumers. This is layer 1 of 2 in native stack #5324; the first
product adoption is
[PR #5323](https://github.com/uzh-bf/klicker-uzh/pull/5323).

This layer intentionally registers no active product flag and initializes no
application. It establishes the reusable boundary first so frontend and backend
rollouts share flag keys, actor attributes, failure behavior, and documentation.

ClickUp Task: N/A (conversation-driven feature spike)

## Proposed Changes

- Add `@klicker-uzh/feature-flags` with pinned GrowthBook SDK `1.6.5` dependencies.
- Add strict flag/default contracts and shared actor attributes (`id`, actor type,
  role, and environment); emails and other direct identifiers are excluded.
- Add a React adapter that creates one browser client per provider mount, updates
  actor attributes without recreating it, and deduplicates initialization.
- Add a process-level Node adapter that evaluates every request with request-local
  attributes, avoiding cross-user context mutation in concurrent services.
- Fail closed to `false` when configuration or the SDK payload is unavailable.
- Document public browser HTTPS versus cluster-internal backend connectivity in
  the feature-flag wiki and ADR 0005.

## Configuration and Security

- Browser adopters use `NEXT_PUBLIC_GROWTHBOOK_API_HOST` and
  `NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY` with a CORS-enabled HTTPS SDK endpoint.
- Backend adopters use `GROWTHBOOK_API_HOST` and `GROWTHBOOK_CLIENT_KEY` with the
  cluster-internal SDK service.
- Client keys identify SDK connections; no GrowthBook management key or secret is
  introduced in this PR.
- Flags are rollout controls, never authentication or authorization boundaries.

### Automated Verification

- [x] `pnpm --filter @klicker-uzh/feature-flags test` — 13/13 tests, including
  browser/Node per-user targeting, deduplicated browser initialization, missing
  configuration, and unavailable SDK payloads.
- [x] `pnpm --filter @klicker-uzh/feature-flags check`
- [x] `pnpm --filter @klicker-uzh/feature-flags build`
- [x] `pnpm run check:all`
- [x] `pnpm run format:check`
- [x] `pnpm run check:syncpack`
- [x] `pnpm run build` — full monorepo production build; also rerun by the
  successful pre-push hook.
[...]

PR #5126, opened 2026-06-17
feat(chat): add versioned chat-engine foundation

rschlaefli:
## What This Adds

- A canonical ordinal `v1` chat-engine contract, fixtures, and black-box conformance runner in the public repository.
- A public default engine using the current AI SDK stack, with strict request validation and one terminal `finish`, `abort`, or `error` event.
- An authenticated `chat-api` tracer for existing threads that retains public ownership of history, persistence, credits, model policy, and trace context.
- Exact deployment-configured origin allowlists before request-scoped provider credentials can cross the host or engine boundary.
- Fail-closed MCP handoff: tools and their scoped execution token are supplied together or no tool is exposed.

The existing Next.js chat route remains the active path. This draft does not cut over the frontend, remove old routes, deploy the new services, or merge the unfinished `v3-ai` roll-up. The embedded assistant and MCP servers on `v3-ai` become later consumers of this seam when that roll-up is ready; they are not prerequisites for this foundation.

## Branch Coverage

- Base: `v3` at `5264353ff77afc598ea69f05f262b25f882ca38c`
- Head: `375d50f290`
- Canonical contract commit used by Catalyst: `44306fb0f2fa3d1d7de7606699232f9247268009`
- Scope: 25 commits and 48 changed files; substantive implementation is 6,021 additions and 9 deletions across 44 files after excluding the lockfile and project artifacts
- The pre-reconstruction source is preserved in local and private archive refs documented by the plan

## Verification

- Node 24: contract 8 tests, default engine 5 tests, and `chat-api` 19 tests passed; all three packages passed focused type-check and build gates.
- Helm charts linted for staging and production, and rendered the exact configured provider origins.
- The unmodified public suite passed against the compiled Catalyst engine with deployment credentials, request credentials, approved tool execution, trace headers, and real transport cancellation. The four terminal results were `finish`, `finish`, `finish`, and `abort`.
- Final security, maintainability, and integrated reviews found no remaining issue at confidence 75 or higher. The repository pre-push build also completed 25 of 25 tasks; the host used Node 22 while the focused evidence used the required Node 24.
- On the implementation head, GitHub type, format, lint, syncpack, build-and-compile, Gitleaks, GitGuardian, and Greptile checks passed. Two install jobs failed when the pre-existing `sharp@0.32.6` binary download returned HTTP 503 and compilation then lacked system libvips; this occurred before branch checks ran.

## Security Boundary

[...]

PR #5421, opened 2026-08-16
fix(feature-flags): harden flag contracts and renumber the ADR

rschlaefli:
Targets `feat/growthbook-foundation` (#5322). Addresses the blockers from an external review of the GrowthBook stack that survived verification against this repository, and records the ones that did not.

## What this changes

| Change | Why |
| --- | --- |
| Invalid non-empty environments normalize to `unknown`; browser and Node adapters then skip the SDK fetch and initialize an empty payload | An environment typo must fail closed even when an ID-only rule or remote default would otherwise enable the flag |
| Deployment environment moves from caller attributes into required adapter configuration | The adapter normalizes it once and overrides any caller-supplied `environment`, keeping request attributes limited to actor identity/context |
| `FEATURE_FLAG_DEFAULTS` typed `satisfies Record<string, false>` | The registry is documented as the safe fallback but is never read at evaluation time — `isOn`/`useFeatureIsOn` resolve unavailable flags through GrowthBook's own `false`. Constraining the registry to `false` makes the two provably agree, and makes a `true` default a compile error rather than a silent lie |
| ADR `0005` → `0008`, with all nine references updated | `0005` is claimed concurrently by #5126 and #5390. `0008` is the first number free against `v3` and every open PR |
| `NEXT_PUBLIC_ENV` added to `turbo.json` `globalEnv` | It reaches the Manage image as a Docker `ENV` rather than through `.env.production`, so it is the one public build variable absent from the Turborepo cache key |
| `docs/feature-flags.md`, new `docs/log/2026-08-16-…` entry | Repo convention: behavior changes update the wiki in the same PR |

Verified on Node 24: **19/19 feature-flag tests pass**, the feature-flags package typecheck/build pass, `pnpm run check:all` passes, the production monorepo build passes (**23/23 tasks**), and Opengrep reports **0 findings**. The invalid-environment regression cases cover an ID-targeted rule, a remote default of `true`, and zero SDK fetches.

## Chose the smaller fix for the defaults registry

The alternative is switching evaluation to `useFeatureValue`/`getFeatureValue` seeded from the registry, which would make a `true` default actually work. That changes evaluation semantics on a foundation PR, so this PR takes the compile-time constraint instead. If you would rather have real registry-driven fallbacks, say so and the swap is small — `docs/feature-flags.md` now names it as the required step before any flag can default on.

## Deployment blocker — not fixed here, and the review's proposed mechanism is wrong

The review says Manage cannot receive `NEXT_PUBLIC_GROWTHBOOK_*` without new Dockerfile `ARG`/`ENV` pairs plus workflow `build-args`. That is not how public variables reach this bundle.

Both image workflows run `mv apps/frontend-manage/.env.stg apps/frontend-manage/.env.production` **before** the Docker build (`v3_frontend-manage-docker-stg.yml` lines 34-38 and 82-86; the prd workflow lines 26-30 and 74-78). Next.js reads `.env.production` off disk during `next build` and inlines the values. That is how all ~20 existing public variables get in. The `ARG NEXT_PUBLIC_*` lines already in `apps/frontend-manage/Dockerfile:31-43` are dead — only `NEXT_PUBLIC_ENV` is promoted to `ENV`, the rest never reach the build.

**So the fix is two lines in `apps/frontend-manage/.env.stg` and `.env.prd`, not a Dockerfile or workflow change.**

It is deliberately not in this PR, for two reasons: the provider that consumes them lands in #5323, not here; and the client key's value is a call for you. It is a public SDK identifier by design and ends up in the browser bundle either way, but this repository is public, so committing it to git history is a decision the maintainers should make explicitly rather than inherit from a fix PR.

## Review claims that did not survive verification

- **"`NEXT_PUBLIC_ENV` must be in `turbo.json` `globalEnv` or staging builds mislabel themselves as production."** `globalEnv` has 107 entries and zero `NEXT_PUBLIC_*` variables, on both `v3` and this branch, yet production bundles embed correct API URLs today. Registration is a cache-invalidation concern, not the injection mechanism. I added the entry for cache correctness and repo convention — it does not fix a mislabelling bug, and should not be described as doing so.
- **"The GrowthBook host and client key are already registered in `globalEnv`."** Neither is present.
- **"Both PRs remain drafts."** #5322 is not a draft.
- **"Cache semantics contradict the fail-closed claim."** The wiki and ADR already state that a normal flag change is picked up on the next provider mount or reload and that GrowthBook may retain a valid cached payload. Nothing here describes an instantaneous kill switch. Invalid deployment environments are handled separately: the adapters never fetch and evaluate an empty payload.

## Open items for the owner

1. **Whether the staging bundle actually embeds `staging`** is still unverified. It cannot be settled by reading — it needs a built staging image inspected for the value. Highest-consequence item in the review and unaffected by this PR.
2. **Red CI on #5322** still needs triage. Note that Playwright shards 5 (`G-elements-mc`/`U-catalog`) and 8 (`N-course` auth-redirect race) are known flakers in this repo; a failure there is not evidence of a regression. Shards 1 and 4 are not on that list, so that is where log-reading pays off.
3. **`docs/adr/README.md` will conflict when you rebase.** This branch is ~63 commits behind `v3`, whose README has since gained `0006` and `0007` rows. The resolution is purely additive: keep the `v3` rows and this PR's `0008` row.
4. **#5390 claims both `0005` and `0006`**, and its `0006` collides with an ADR already merged to `v3`. Out of scope here, but it is the same failure and worth a heads-up to that author.
5. **SDK pin left at `1.6.5`** although `1.7.0` shipped 2026-08-07. Deliberate: bumping it would add an untested dependency change to a PR whose purpose is unblocking. Worth a separate PR.
6. **#5323 needs a small rebase update after this PR lands:** pass `NEXT_PUBLIC_ENV ?? NODE_ENV` through the required client `environment` config and keep only `id`, `actorType`, and `role` in actor attributes.

Opened as a draft so it does not trigger the gated image-build jobs. Mark it ready whenever you want the full matrix.


jabbadizzleCode (approved):
LGTM. Added the setting to prevent feature flags from activating when the deployment environment is invalid.

PR #5837, opened 2026-09-08
fix(manage): keep block actions clear of scrollbars

pascalr77:
## Description

Keep the scrollbar in activity creation/editing element lists from covering the trash and reorder buttons. The shared `WizardElementList` now reserves 16px of space on the right while retaining its existing scrolling behavior.

ClickUp Task: [Block list scrollbar overlaps the trash button in Chromium and WebKit](https://app.clickup.com/t/4648007/123zgec18rd)

## Proposed Changes

- Add right padding to the shared block/stack element list.
- Add a Playwright regression covering overflowing lists with and without a pending selection, button clearance and hit testing, reordering, and deletion after scrolling. Register it with CI's Manage-only profile and relevance manifest.
- Allow the host test launcher to select `--profile manage` and install Chromium with `--install-browser` without reconciling containers or reinstalling workspace dependencies. Add launcher coverage and usage notes.

## How to Test & Verification Evidence

```bash
node util/run-playwright-host.mjs --profile manage W5-block-list-scrollbar.spec.ts --project chromium
```

Add `--headed` for visible Chromium or `--ui` for Playwright UI mode.

### Automated Verification

- Chromium headless: 2 passed (cleanup + regression).
- Chromium headed: 2 passed in 7.8 seconds.
- Negative control: removing the padding fails the button-clearance assertion; the fix was restored.
- Host launcher tests: 27 passed.
- CI selector and runtime-profile tests: 19 passed on the host after registering the spec.
- Formatting and diff whitespace checks passed.
- `pnpm run check:all` was attempted in the container but could not finish: the runtime-profile test requires a host devrouter executable. The spec-registration failures discovered in that run were corrected and the affected suite passed on the host.
- Full `pnpm run build`: passed, 23/23 tasks (3m41s).

### Manual/Visual Verification

Checked the activity wizard with delegated synthetic credentials using agent-browser. With ten rows, the list retained a 120px viewport over 289px of content and 16px of right padding. Captured before/after screenshots locally; the before capture removes the added padding only in the browser DOM.

WebKit is not verified locally because of the local Safari login loop. A reported UI-mode timeout was not reproduced in the headed run; its cause remains unresolved. Keeping this PR in draft pending broader verification.

### DIFF
--- a/playwright/tests/W5-block-list-scrollbar.spec.ts
+++ b/playwright/tests/W5-block-list-scrollbar.spec.ts
@@ -1,0 +1,112 @@
+import { expect, type Locator } from '@playwright/test'
+import { cleanupTest } from '../util/cleanup.js'
+import { USER_ID_TEST } from '../util/constants.js'
+import { test } from '../util/fixtures.js'
+import { createQuestionSC } from '../util/fixtures/elements.js'
+
+test('CLEANUP', cleanupTest)
+
+// Overlay scrollbars are platform dependent and may be hidden in headless
+// browsers. Check physical clearance as well as clickability: a successful
+// click alone would also pass on the original, overlapping layout.
+async function expectScrollbarClearance(row: Locator) {
+  await row.scrollIntoViewIfNeeded()
+  await expect
+    .poll(() =>
+      row.evaluate((element) => {
+        const list = element.parentElement!
+        const bounds = list.getBoundingClientRect()
+        return Array.from(element.querySelectorAll('button')).every(
+          (button) => {
+            const rect = button.getBoundingClientRect()
+            const hit = document.elementFromPoint(
+              rect.x + rect.width / 2,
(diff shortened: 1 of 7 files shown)


playwright/tests/W5-block-list-scrollbar.spec.ts
```diff
@@ -0,0 +1,112 @@
+import { expect, type Locator } from '@playwright/test'
```
rschlaefli, line 1: @pascalr77 I would remove this test as its on a very granular level (almost like a unit test) and we would want to keep Playwright for E2E testing of the key user flows. If we do a lot of testing like this our test runtime will explode even more. Maybe you could add/improve AGENTS.md to make sure such tests are not generated but rather and maybe added to an existing test flow that actually passes through this functionality.

rschlaefli (changes requested):

PR #6054, opened 2026-09-15
chore(integration): merge latest v3 into v3-audit

rschlaefli:
## What

Merges `v3` (`cfa279ab22`, #6041) into `v3-audit`, the staging source branch.

## Why

`v3-audit` is what staging builds from (`STG_SOURCE_BRANCH=v3-audit`). It was still carrying the red `build-images-status`, `scan-arm` and `scan-migrator-arm` lanes because the image-scan remediation had only reached `v3`. #6051 brought it to `v3-ai`; this carries it onto `v3-audit`.

## Relation to #6052

#6052 forward-merged the `v3-ai` chat fix (#6044). It deliberately carried no v3 content, so the remediation still had to land here. This branch was refreshed onto `109c7d96a4` (#6052's merge) before opening, so it contains both.

## Content

Exactly `cfa279ab22`'s 24 paths:

- runtime base images to `node:24.21.0-alpine` with `apk upgrade`
- `pnpm`/`turbo` removed from the runtime stages in the same layer as the install
- range-scoped `overrides` in `pnpm-workspace.yaml` for the fixable HIGH/CRITICAL transitives

## Resolution notes

- `pnpm-lock.yaml`: kept the `v3-audit` side (so the `packages/audit` and `packages/logging` entries survive) and regenerated with the pinned `pnpm 11.5.0`. Verified with `pnpm install --lockfile-only --frozen-lockfile`; worktree clean.
- `docs/data-and-migrations.md`: keeps v3-audit's staging-revision text and takes v3's migrator-image update.

## Not covered

SonarCloud is red on `v3` itself and is inherited here; this merge does not clear the staging promoter's SonarCloud requirement.

PR #6056, opened 2026-09-15
chore(integration): synchronize v3-ai into v3-audit

rschlaefli:
## What

Forward-merges `v3-ai` into `v3-audit`, the branch staging builds from (`STG_SOURCE_BRANCH=v3-audit`).

## Why

This is the only path for v3 content to reach `v3-audit`: v3 fixes land on `v3-ai` first, then merge forward here. An earlier attempt (#6054) merged `v3` straight into `v3-audit` and was closed for that reason.

`v3-audit` was still red on `build-images-status`, `scan-arm` and `scan-migrator-arm` because the image-scan remediation had only reached `v3`. #6051 carried it to `v3-ai`; this brings it onto the staging branch.

## Content

26 files over `v3-audit`:

- `cfa279ab22` (#6041)'s 24 paths: runtime base images to `node:24.21.0-alpine` with `apk upgrade`, `pnpm`/`turbo` removed from the runtime stages in the same layer as the install, and range-scoped `overrides` in `pnpm-workspace.yaml` for the fixable HIGH/CRITICAL transitives.
- #6053's `util/local-kb` provider failure-code fix, which was on `v3-ai` but not yet here.

## Resolution notes

- `pnpm-lock.yaml` was the only conflict. The `v3-audit` side was kept so the `packages/audit` and `packages/logging` entries survive, then regenerated with the pinned `pnpm 11.5.0`. `pnpm install --lockfile-only --frozen-lockfile` passes and the worktree is clean; the lockfile also passes the supply-chain policy check.
- `docs/data-and-migrations.md` auto-merged, keeping v3-audit's staging-revision text together with v3's migrator-image update.

## Not covered

SonarCloud is red on `v3` itself and is inherited here, so this merge does not clear the staging promoter's SonarCloud requirement.


rschlaefli:
Closing as a duplicate of #6055. Both merge v3-ai into v3-audit and touch the identical 26-file set; #6055 was opened first and its branch carries no incidental commits, so it should be the one that lands. No v3 content reached v3-audit directly.

PR #6088, opened 2026-09-16
enhance(kg): anchor per-slot evidence and scope the content digest to course content

rschlaefli:
## Problem

Two coupled defects in element generation, plus one graph-staleness defect this branch also carries from its baseline.

**1. Every generated question in a bank carried the same objective and the same evidence pages.** When a design slot carried any learning objective, the worker never used that slot's independently selected graph concept as an evidence anchor, so every module-scoped slot in a module received the same module-wide top-ranked bundle. A live PRD build (`c7370d09`) assigned six distinct concepts across six slots and produced identical evidence entities `[Gesamtkapitalebene, Entity Approach, DCF-Methode, Ertragswert-Methode]`, identical pages `[182, 48]`, and six identical question titles.

The fabricated **neutral** objective is a second cause of the same collapse. Element generation synthesizes a Bloom-level placeholder objective when the lecturer has not written one, and that placeholder is the only carrier of the lecturer's Bloom-level choice into the worker blueprint. Because every slot in a run shares the placeholder text, it biased retrieval identically across slots.

**2. Administrative uploads marked a current knowledge graph stale.** The staleness digest counted every resource serving RAG with no `materialType` filter, so re-uploading a syllabus flipped a current graph to stale. That flag is not cosmetic: `apps/chat` hard-blocks graph retrieval on `isStale`, so a current graph stopped answering after an administrative upload. On the live Finance I HS26 knowledge base the all-resources digest was `d02c2542b37d5743…` while the stored digest of published build `d257187f` was the course-content digest `00f7f6acfd9e3dc6…` — the build was current and the comparator was wrong.

## Result

- Each slot's own selected concept anchors its evidence ranking, so slots in one module resolve different evidence bundles.
- Slots whose objective is a synthesized placeholder are marked and treated as objective-less for retrieval, so the placeholder no longer suppresses the anchor.
- The design review shows the evidence entities each slot resolved and raises an advisory concentration notice when a module's slots share one identical set.
- The staleness digest is scoped to course content, mirroring the graph build path.

## How it works

**Neutral-objective marker (Klicker side).** Synthesized Bloom-level objectives are marked `objectiveSource: 'neutral'`; lecturer-authored objectives are `'provided'`. The marker is exposed on the `ElementGenerationObjective` GraphQL view and carried through the configuration, design-summary, and UI objective types. A build persisted before the marker has no value and is treated as lecturer guidance, so only an explicit `'neutral'` shows the generated-default label in the review gate. This is deliberate: marking by absence would misclassify every existing build.

**Emission gate (important for rollout).** The `objective_source` key is **not** written into the worker blueprint JSON in this branch. It is gated behind `BLUEPRINT_OBJECTIVE_SOURCE_ENABLED = false` in `packages/graphql/src/services/questionGenerationBlueprint.ts`. The currently deployed PRD worker validates blueprint objective rows against a closed allowlist and rejects unknown fields, so shipping the key early would break production generation. Enabling it is a one-line change that must follow deployment of the companion worker release.

**Per-slot evidence surface.** `parseQuestionGenerationDesign` reads the evidence entity ids the worker already persists at `resolved_slots[].graph_resolution`, preferring `evidence_candidates[0].entity_ids` with an `entity_ids` fallback, and deduplicates them in order. Only ids leave the server: the rest of `graph_resolution` (chunk ids, paths, scores, citations) stays behind a `.passthrough()` schema and never reaches a GraphQL view or the UI. The existing internals-leak guard test is retained and extended.

**Concentration notice.** The review gate lists each slot's surfaced entities and warns when a module's slots resolve one identical set. Slots without surfaced entities render nothing extra, so pre-marker builds and older worker artifacts stay valid. The view-model logic lives in a dependency-free module (`designReviewSummary.ts`) rather than in the component.

**Digest scope.** `readKBContentDigestEntries` now filters to `KBResourceMaterialType.COURSE_CONTENT`, mirroring the build path in `packages/graphql/src/services/knowledge.ts`. Existing builds gain no rows; only the staleness comparison changes.

## Compatibility and scope

- No schema, migration, dependency, environment-variable, or API-breaking change. `schema.graphql` is regenerated for the added views and field.
- Existing persisted designs, blueprints, and builds remain valid; every new field is optional or has a defined legacy default.
- Source-only. No deployment, image publication, or cluster change is included.

## Verification

- `packages/knowledge-graph`: 80 passed, 2 skipped (the two integration suites need a live FalkorDB). `tsc --noEmit` clean for that package.
- `packages/graphql` focused suites (`questionGenerationArtifacts`, `questionGenerationBlueprint`, `questionGenerationConfiguration`): 114 passed, including the emission-gate tests in both directions and the retained internals-leak guard.
- `apps/frontend-manage` `node --test` suite: 10 passed, including the new evidence and objective-source tests. Biome clean on all changed files.
- `apps/frontend-manage` typecheck: 35 errors, none in the changed files. The baseline was established rigorously by temporarily restoring the pre-change component and view-model pair and diffing sorted error lists — the 35 are pre-existing and come from unbuilt `@klicker-uzh/markdown` / `@klicker-uzh/word-cloud` workspace dependencies and shared-components JSX noise. The two errors this branch originally introduced (the view model widened the Bloom level to `string`, breaking the fragment assignment and the namespaced message key) are fixed.

## Not covered / rollout blockers

- **The worker-side half is not in this repository.** The companion change in `kg-content-generation` (draft MR !20, branch `fix/question-evidence-slot-anchoring`) accepts and propagates the marker and implements per-slot anchoring. The emission gate must stay off until that release is merged, published, and deployed to PRD.
- **No live end-to-end proof.** The broker-backed generation path and a real PRD rebuild are deliberately out of scope here and belong to the separately authorized deployment and e2e sequence.
- **No browser capture of the changed review surface.** The review gate gains the evidence list, the concentration notice, and the generated-default label, but this package ran no browser session, so the rendered result is unverified. The view-model logic is covered by unit tests on the rendering branches; the actual layout and both locales still need a real capture before this is called UI-verified.
- **The `ocr-review` check fails on infrastructure, not content.** All 13 selected items failed with `provider or subtask request failed` and `Unable to read the installed OpenCodeReview version`; 0 findings were produced. It needs a re-run.
- The `final-ai-review` check is pending and requires a manual review dispatch.


apps/frontend-manage/src/components/elements/generation/designReviewSummary.ts
```diff
@@ -0,0 +1,93 @@
...
+    text: string
+    bloomLevel?: ElementGenerationBloomLevel | null
+    objectiveSource?: string | null
+  }>
+): ElementGenerationDesignObjectiveView[] {
+  return objectives.map((objective) => ({
+    id: objective.id,
+    text: objective.text,
+    bloomLevel: objective.bloomLevel ?? null,
+    isGeneratedDefault: objective.objectiveSource === 'neutral',
```
rschlaefli: maintainability · medium
The generated-default flag is derived from an inline magic literal ('neutral') while the parameter is typed as a bare string. This duplicates the backend contract value that already has an authoritative union type (QuestionGenerationObjectiveSource in @klicker-uzh/types) without any compile-time link: a typo such as 'neutal' would type-check, and a backend rename of the marker would silently stop rendering the label. Widening the input type also loses the union guarantee at the call site. Use the shared union for the input parameter and a named constant for the marker instead of the raw literal.

**Suggestion:**
```suggestion
    bloomLevel: objective.bloomLevel ?? null,
    isGeneratedDefault: objective.objectiveSource === NEUTRAL_OBJECTIVE_SOURCE,
```

---

Verified and fixed in `d089967628`.

`NEUTRAL_OBJECTIVE_SOURCE` is now the single marker constant, `designReviewObjectives` takes `QuestionGenerationObjectiveSource` instead of `string`, and the generator normalizes with the same constant. The deeper cause was the GraphQL view using `exposeString`, which erased the union from the generated client types and let the raw literal reach this file: `objectiveSource` is now a real `ElementGenerationObjectiveSource` enum. Regenerated `schema.graphql` and the client documents. Re-ran the affected suites — graphql 140/140 (configuration, artifacts, lifecycle, blueprint, accounting), frontend-manage 10/10, `tsc` clean for graphql/types/frontend-manage, `format:check` clean.

---

⚠️ **Correction from the agent:** this finding was accidentally deleted from its original position by an errant API call while replying, so its text is reproduced above verbatim. Nothing was dismissed. The thread is resolved because the fix landed, not because the finding was withdrawn.

PR #6214, opened 2026-09-21
fix(ci): follow the registry blob redirect when reading image revisions

rschlaefli:
## Problem

`Promote to stg` run [35549792165](https://github.com/uzh-bf/klicker-uzh/actions/runs/35549792165) failed at 01:16:54Z with

```
[TypeError: fetch failed] { [cause]: Error: unexpected redirect at makeNetworkError (undici) }
```

Candidate `6ecaac08ebd1` (#6212) never reached `stg-release`, so staging still runs the previous release.

## Root cause

GHCR serves a manifest where it is addressed, but a blob only through a 307 redirect to `pkg-containers.githubusercontent.com`. `fetchImageRevision()` reads the image config blob to resolve the revision of an adopted digest and used `redirect: 'error'`, so undici aborted with `unexpected redirect`. Reproduced against the real registry: manifest 200, config blob 307 to the storage host.

Roughly half the promoted images are adopted digests (`image-reuse-*.json` with `adopted: true`), so a normal promotion hits this path.

## Change

- blob reads use `redirect: 'manual'`; exactly one hop is followed, only to `pkg-containers.githubusercontent.com` over https; manifests still fail fast on any redirect
- the storage response is accepted with `application/octet-stream`, and the bytes must hash to the requested digest
- the signed storage URL is fetched without the registry token, so the token stays with the registry
- any further redirect fails

## Verification

- `node --test .github/scripts/stg-release-promoter.test.js` → 50 pass / 0 fail (3 new tests: follows one hop, rejects a hop off the storage host, rejects substituted bytes)
- real registry read of candidate `6ecaac08ebd1`: 10/10 image targets resolved, including `analytics-arm`, `lti-arm`, `mcp-lecturer-arm` and `olat-api-arm` adopted from an earlier build (`ddfa6ee4d757…`) plus `chat-arm` at the candidate revision
- `biome format` and the file's baseline lint warnings unchanged on both files

## After merge

The `promote` job only runs for a workflow that completed on the source branch (`v3-audit`), so the wake-up comes from the next push there: `v3` is merged forward through `v3-ai` into `v3-audit`, and the resulting candidate CI plus image build re-evaluates the promotion. No manual dispatch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved processing of adopted images that use registry storage redirects.
  - Securely supports a single approved redirect while rejecting redirects to untrusted hosts.
  - Validates downloaded image data against its expected integrity hash.
  - Rejects unexpected content types and invalid image data, helping prevent incorrect or unsafe release processing.


rschlaefli:
## On the HIGH finding `fr-1c336d180c476fee`

The finding assumes undici returns the browser's opaque redirect for `redirect: 'manual'`. On the runtime the promoter uses it does not. Probe on Node v24.17.0 (the `node24` line of `actions/github-script` v9):

```
$ node -e "const r = await fetch('https://httpbingo.org/redirect/1', {redirect:'manual'}); console.log(r.status, r.type, r.headers.get('location'))"
302 basic /get
```

End-to-end: reading an adopted digest's revision through this function resolved all 10 staging image targets of candidate `6ecaac08eb`, including `analytics-arm`, `lti-arm`, `mcp-lecturer-arm` and `olat-api-arm` adopted from an earlier build. That path only completes if the 307 from `ghcr.io` to `pkg-containers.githubusercontent.com` is visible and followed, and the returned bytes hash to the requested digest. Were a future runtime to return the opaque redirect, the code fails closed on `registry response was 0`; the guard for substituted content is the digest hash.

PR #6248, opened 2026-09-21
fix(dpo): gate response collection on account data-use completion

rschlaefli:
### Problem

`v3.4.0-alpha.81` collects live-quiz and assessment responses without checking whether the responding participant account has completed the current data-use disclosure. An account that never acknowledged the disclosure, or that is still on the previous disclosure version, can therefore have responses collected, scored and audited. The GraphQL surface already refuses such accounts, so the response API is the remaining ingress that can accept data the participant never agreed to.

### Result

Both response-API handlers now apply the same completion predicate that the GraphQL gate uses, before any event is pushed:

- `POST /` (regular live quiz) rejects a request whose `participant_token` verifies as a `PARTICIPANT` account that has not completed the disclosure, answering `403 {"error":"PARTICIPANT_DATA_USE_COMPLETION_REQUIRED"}` and enqueueing nothing. Anonymous responses and responses that carry only `temporary_participant_token` are untouched, so live quizzes keep working for participants without a registered account.
- `POST /assessment` rejects every submission from an account that has not completed the disclosure, because assessment responses are always attributed. The rejection is recorded as an audit-log entry alongside the existing validation failures.

The PWA question area maps HTTP 403 to a new toast, stores the current question URL as the return target, and redirects to `/account/data-use`, mirroring the existing GraphQL-gate behaviour.

### Design

`apps/response-api/src/participantAdmission.ts` holds the decision as a pure module so the branching is testable without an HTTP server: `admitRegisteredParticipant` for the always-attributed assessment path, and `admitParticipantToken` for the regular path, which reads only the `participant_token` cookie. A cookie that does not verify, or that does not resolve to a registered participant, is admitted unchanged: those requests were already treated as anonymous before this gate, and rejecting them would break live quizzes for participants holding a stale cookie without protecting anything.

The predicate itself is `isParticipantDataUseComplete` from `@klicker-uzh/util`, so the response API and the GraphQL gate cannot drift apart on what "completed" means: the acknowledgement must carry `PARTICIPANT_DATA_USE_DISCLOSURE_VERSION` and both optional choices must have been recorded with their disclosure version. No new migration or schema change is involved.

The gate needs database access, so the chart now supplies `DATABASE_URL` to both response-api deployments from the shared `-secret-backend-graphql` secret, the same secret the GraphQL backend uses. `APP_SECRET` was already mounted for the assessment cookie verification.

### Compatibility and scope

`403` was previously unused by this service, so the PWA mapping cannot misread an existing failure. Temporary-participant and anonymous collection, scoring, gamification and audit behaviour are unchanged. This layer does not add exports, does not change Learning Analytics computation, and does not change what is released or displayed after collection.

### Verification

- `pnpm --filter @klicker-uzh/response-api run check` and `run test:run` (10 node:test cases in `participantAdmission.test.ts`), plus `pnpm --filter @klicker-uzh/frontend-pwa run check`.
- Runtime proof against the containerised response API: incomplete registered account → `403` with zero events enqueued; anonymous request → `200` with one event; completed account → `200` with one event. Assessment mode on a temporary instance: incomplete → `403`, completed → `200 {"status":"response_submitted"}`.
- Browser proof of the contract the gate reads: a seeded participant is redirected to `/account/data-use`, records the choices, and the database read-back satisfies the gate predicate.
- Helm rendering for `env-uzh-prd` and `env-uzh-stg`: both response-api containers now receive `DATABASE_URL` from the backend secret.
- Container gates `check`, `check:lint`, `check:syncpack`, `check:agents-md`, `check:git-identity`, `check:removed-doc-artifacts` and `check:prisma-sync` pass. `check:playwright-ci` cannot run outside a host Devrouter context and is covered by CI instead.
- Browser proof of the client reaction in a running live quiz: with the participant on an active question and the account carrying a stale disclosure version, submitting the answer shows the new toast and redirects to `/account/data-use`; after restoring the current version the same submission is accepted ("Your answer has been successfully submitted and saved.") and the quiz advances to the next question.

### Screenshots

| State | Capture |
| --- | --- |
| Live-quiz answer rejected for a stale disclosure version |  |

The capture shows the toast and the redirect to `/account/data-use`. It was taken in the containerised stack on `rs/dpo-response-admission` with the synthetic participant's disclosure version moved to a stale value, reproducing a policy update during an open session.


rschlaefli:
Closing unmerged. The response-ingress gate is dropped from the DPO release.

**Decision.** No data-use check runs on the response hot path. The disclosure
stays enforced where it binds today: `packages/graphql/src/lib/participantAccountGate.ts`
wraps every GraphQL root field for registered participants, so an account that
has not confirmed the current disclosure cannot load a quiz or an assessment,
and the PWA redirects it to `/account/data-use`. The plan's framing is a use
boundary: operational responses, scoring, assessment audit logs and chat history
stay available to serve the request, without being treated as optional
learning-analytics derivatives. Under that framing no ingress check is required.

**Cost of the gate.** It added a Prisma read per submission for every request
carrying a registered participant token, in both response APIs, plus a new
`DATABASE_URL` dependency for those containers (#6250 is its companion). It also
rejected legitimate mid-quiz traffic: a valid-token participant whose account had
not yet confirmed the disclosure got a 403 at submission time. The current
ingest path reads no database at all; token verification happens in the worker,
which nulls an unverifiable participant and lets the answer continue as an
anonymous one.

**Consultation.** GLM 5.3 Flash reviewed the design twice and recommended
dropping the gate together with the deployment change. It corrected one claim in
the earlier rationale: the response API does not verify the JWT on the live path.

**Residual accepted.** Valid-token submissions from not-yet-completed accounts
continue on the live path (pre-bump sessions, scripted posts). If a later DPO
framing requires collection-time evidence, it belongs in a deferred layer as a
version-stamped completion flag or an append-only consent-history event, and
never as a per-response read.

**Scope of the drop.** `02231787ef` stays on this branch for reference; nothing
from it lands on `v3`. The gate helper exists on this branch alone, and `v3`,
`v3-ai` and every follow-up layer are free of it, so no other branch needs a
matching revert. This deliberately diverges from the earlier integration branch
`fc59e64f9d`, which carried the same inline read.

PR #6250, opened 2026-09-22
chore(deploy): provide the response-api database URL

rschlaefli:
## Problem

The response APIs now gate collection on the participant's data-use state, so each response-api container needs the database URL that state lives in. Neither the regular nor the assessment container carries one: the response-api secret supplies the Hatchet token, the Redis passwords, and `APP_SECRET` only.

Without the credential the gate cannot read the account and every gated response is rejected, so both halves of the change have to reach production together.

## Change

`deploy/charts/klicker-uzh-v3/templates/deployment-response-api.yaml` gives both containers `DATABASE_URL` from the externally provisioned `…-secret-backend-graphql` secret, matching the migrator hook and the lecturer MCP deployment. The assessment stack shares that database, verified empirically on staging by probing `checkParticipantNameAvailable` against `api` and `assessment-api`.

This is the `v3-ai` half of the deploy-parity requirement. The companion change ships with the response-api admission gate on `v3` ([#6248](https://github.com/uzh-bf/klicker-uzh/pull/6248), `fix(dpo): gate response collection on account data-use completion`). Production renders `deploy/` from `v3` while the release images are tagged from `v3-ai`, so both branches must carry the same `deploy/` revision; #6248's `check`/`check-suite` contexts are red until this revision exists on `v3-ai`.

## Verification

`node .github/scripts/deploy-parity.cjs --other rs/dpo-response-api-db-url --candidate HEAD` in the #6248 worktree reports `deploy parity ok` against this branch, i.e. the two revisions are identical under `deploy/` apart from the environment-owned image reference lines.

The env var is inert on this branch until the gate source arrives through the ordinary `v3` → `v3-ai` integration.

Draft: no merge implied.


rschlaefli:
Closing unmerged. The response-ingress gate that needed this deployment change is
dropped from the DPO release, so both response API containers keep their current
secret set.

#6248 records the decision and the full rationale: the disclosure is enforced at
the GraphQL account gate, the ingest path reads no database today, and the gate
would have added a per-submission read plus a mid-quiz rejection for accounts
that have not confirmed the disclosure yet. GLM 5.3 Flash reviewed the design and
recommended dropping both halves.

